PR #33, opened 2026-09-13
T02 Core domain model and JSON round-trip

### LINKED ISSUE #2: T02 Core domain model
**Design milestone:** M1 (types half) · **Branch:** `task/T02-domain-model`
**Model / effort:** **Opus / High** · **Reviewer:** Opus / High **+ `/code-review --effort ultra`**
**Owns:** `src/IC2.Engine/Model/**`, `src/IC2.Engine/Serialization/**`, `data/worlds/toy-3city.json`, `data/rulesets/toy-ruleset.json`, `data/scenarios/toy-3city.json`, `tests/IC2.Engine.Tests/Model/**` (file-level, **not** `data/worlds/**` / `data/rulesets/**` wildcards — T29 owns the real shipped world/ruleset data under the same directories)

### Scope
`World` / `Ruleset` / `Scenario` / `SaveGame` as C# records with `System.Text.Json` contracts, plus the `GameState` tree they produce. Must carry, from day one, every field later tasks need so nobody has to widen the model: per-army and per-fleet **money purse and supply stock**, army morale (record `+14` semantics), the unit slot regular/mercenary marker, city fortification with its `>100` in-progress encoding, fleet condition, the symmetric N×N relation matrix with negative cooldowns, and the news ring buffer's storage. Support a `"_provenance"` key per field. Ship the toy 3-city / 2-nation world + ruleset + scenario under `data/`. Terrain tile types are an **open list**, not hardcoded to the original's 12 codes.

### Done when
- [ ] Round-trip equality tests pass for each of `World`, `Ruleset`, `Scenario`, `SaveGame`
- [ ] The toy scenario loads and resolves its world and ruleset by id
- [ ] A malformed file, an unknown required field, and a version mismatch each produce a distinct typed error — one negative test each, never a silent default
- [ ] `GameState` round-trips: build a non-trivial state, serialize, deserialize, assert deep equality
- [ ] No gameplay constant is hardcoded in C#: a test asserts every ruleset-governed number comes from the loaded `Ruleset`

### Blocked by
- [ ] #1 T01 Build scaffolding and CI

### Hazard
The widest-blast-radius diff in the plan. Nothing else touching `src/IC2.Engine` may be in flight.

---
Full task entry (the binding contract): [`docs/build-orchestration-plan.md` → T02 Core domain model](https://github.com/diegoami/imperial_conquest_2/blob/main/docs/build-orchestration-plan.md#t02-core-domain-model)

diegoami:
Closes #2

## Task

- **Task**: T02 Core domain model and JSON round-trip — [plan entry](https://github.com/diegoami/imperial_conquest_2/blob/main/docs/build-orchestration-plan.md#t02-core-domain-model-and-json-round-trip)
- **Branch**: `task/T02-domain-model`

| --- | --- |
| `src/IC2.Engine/Model/FortificationCode.cs` | `src/IC2.Engine/Model/**` |
| `src/IC2.Engine/Model/GameDataSchema.cs` | `src/IC2.Engine/Model/**` |
| `src/IC2.Engine/Model/GameState.cs` | `src/IC2.Engine/Model/**` |
| `src/IC2.Engine/Model/GameStateFactory.cs` | `src/IC2.Engine/Model/**` |
| `src/IC2.Engine/Model/JsonContract.cs` | `src/IC2.Engine/Model/**` |
| `src/IC2.Engine/Model/Provenance.cs` | `src/IC2.Engine/Model/**` |
| `src/IC2.Engine/Model/Ruleset.cs` | `src/IC2.Engine/Model/**` |
| `src/IC2.Engine/Model/RulesetNumbers.cs` | `src/IC2.Engine/Model/**` |
| `src/IC2.Engine/Model/SaveGame.cs` | `src/IC2.Engine/Model/**` |
| `src/IC2.Engine/Model/Scenario.cs` | `src/IC2.Engine/Model/**` |
| `src/IC2.Engine/Model/UnitSlot.cs` | `src/IC2.Engine/Model/**` |
| `src/IC2.Engine/Model/ValueList.cs` | `src/IC2.Engine/Model/**` |
| `src/IC2.Engine/Model/World.cs` | `src/IC2.Engine/Model/**` |
| `src/IC2.Engine/Serialization/GameDataExceptions.cs` | `src/IC2.Engine/Serialization/**` |
| `src/IC2.Engine/Serialization/GameDataLoader.cs` | `src/IC2.Engine/Serialization/**` |
| `src/IC2.Engine/Serialization/GameDataRepository.cs` | `src/IC2.Engine/Serialization/**` |
| `src/IC2.Engine/Serialization/GameDataValidation.cs` | `src/IC2.Engine/Serialization/**` |
| `src/IC2.Engine/Serialization/GameJson.cs` | `src/IC2.Engine/Serialization/**` |
| `src/IC2.Engine/Serialization/SchemaValidator.cs` | `src/IC2.Engine/Serialization/**` |
| `data/worlds/toy-3city.json` | `data/worlds/toy-3city.json` (file-level) |
| `data/rulesets/toy-ruleset.json` | `data/rulesets/toy-ruleset.json` (file-level) |
| `data/scenarios/toy-3city.json` | `data/scenarios/toy-3city.json` (file-level) |
| `tests/IC2.Engine.Tests/Model/GameStateSerializationTests.cs` | `tests/IC2.Engine.Tests/Model/**` |
| `tests/IC2.Engine.Tests/Model/ModelCoverageTests.cs` | `tests/IC2.Engine.Tests/Model/**` |
| `tests/IC2.Engine.Tests/Model/NoHardcodedConstantsTests.cs` | `tests/IC2.Engine.Tests/Model/**` |
| `tests/IC2.Engine.Tests/Model/RoundTripTests.cs` | `tests/IC2.Engine.Tests/Model/**` |
| `tests/IC2.Engine.Tests/Model/ScenarioLoadingTests.cs` | `tests/IC2.Engine.Tests/Model/**` |
| `tests/IC2.Engine.Tests/Model/TestPaths.cs` | `tests/IC2.Engine.Tests/Model/**` |
| `tests/IC2.Engine.Tests/Model/ToyFixtures.cs` | `tests/IC2.Engine.Tests/Model/**` |
| `tests/IC2.Engine.Tests/Model/TypedLoadErrorTests.cs` | `tests/IC2.Engine.Tests/Model/**` |

30 files, all additions, no deletions, no modifications to any existing file.
`git status` on `main` was clean before the branch was cut; `git diff --stat main...HEAD` shows
only the paths above. In particular: `IC2.sln`, `Directory.Build.props`, `.editorconfig`,
`.github/workflows/**`, `src/IC2.Data`, `src/IC2.Inspect`, `src/IC2.Cli`, `godot/`,
`tests/IC2.Engine.Tests/IC2.Engine.Tests.csproj`, `tests/IC2.Engine.Tests/PlaceholderTests.cs`,
`src/IC2.Engine/Placeholder.cs` and every file under `docs/` are untouched.

- [x] No change to `docs/build-orchestration-plan.md`, original game files, `assets.local.ini`, `.gitignore`'s exclusion policy, or anything under `docs/reports/` (each of those is an escalation, §6.6).

Note on one deliberate non-change: the test project file was **not** edited to copy `data/` into the
test output, because `tests/IC2.Engine.Tests/IC2.Engine.Tests.csproj` is T01's, not this task's.
`tests/IC2.Engine.Tests/Model/TestPaths.cs` instead walks up from the test assembly's directory to
`IC2.sln` and reads the committed files in place — in scope, and it also means the tests assert
against the files that actually ship rather than against a copy of them.

## DoD evidence

```text
Build gate: dotnet build IC2.sln, zero warnings in the new projects
$ dotnet build IC2.sln --nologo -c Debug /t:Rebuild
Build succeeded.
    0 Warning(s)
    0 Error(s)

Test gate: dotnet test IC2.sln
$ dotnet test IC2.sln --nologo -v q
Passed!  - Failed: 0, Passed:  1, Skipped: 0, Total:  1 - IC2.Data.Tests.dll (net10.0)
Passed!  - Failed: 0, Passed: 45, Skipped: 0, Total: 45 - IC2.Engine.Tests.dll (net10.0)

DoD 1: Round-trip equality tests pass for each of World, Ruleset, Scenario, SaveGame
       (serialize -> deserialize -> serialize produces identical JSON).
$ dotnet test IC2.sln --nologo -v q --filter "FullyQualifiedName~RoundTripTests"
Passed!  - Failed: 0, Passed: 5, Skipped: 0, Total: 5 - IC2.Engine.Tests.dll (net10.0)
  (World_round_trips_to_identical_json, Ruleset_round_trips_to_identical_json,
   Scenario_round_trips_to_identical_json, SaveGame_round_trips_to_identical_json,
   World_provenance_entries_survive_the_round_trip_in_order)
  Each asserts the first and second serializations are byte-identical strings AND that the
  reloaded document is == the original, so a field that round-trips to the same text but a
  different value cannot pass.

DoD 2: The toy scenario loads from data/scenarios/toy-3city.json and resolves its world and
       ruleset by id.
$ dotnet test IC2.sln --nologo -v q --filter "FullyQualifiedName~ScenarioLoadingTests"
Passed!  - Failed: 0, Passed: 5, Skipped: 0, Total: 5 - IC2.Engine.Tests.dll (net10.0)
  (Toy_scenario_loads_from_its_shipped_file,
   Toy_scenario_resolves_its_world_and_ruleset_by_id,
   Resolving_a_scenario_produces_a_startable_state,
   An_unknown_scenario_id_is_an_unresolved_reference_not_a_null,
   A_scenario_naming_a_missing_world_is_an_unresolved_reference)
  Resolution is by the id declared INSIDE each file, not by filename: GameDataRepository scans
  data/worlds, data/rulesets and data/scenarios in ordinal filename order and indexes on the
  documents' own ids.

DoD 3: A malformed file, an unknown required field, and a version mismatch each produce a
       distinct typed error, one negative test each - never a silent default.
$ dotnet test IC2.sln --nologo -v q --filter "FullyQualifiedName~TypedLoadErrorTests"
Passed!  - Failed: 0, Passed: 9, Skipped: 0, Total: 9 - IC2.Engine.Tests.dll (net10.0)
  A_malformed_file_throws_MalformedGameDataException                        -> malformed
  A_missing_required_field_throws_MissingRequiredFieldException             -> required field absent
  A_missing_required_field_nested_in_a_list_names_its_full_path             ->   "seats[1].control"
  An_unrecognised_field_throws_UnknownFieldException_rather_than_being_dropped -> field not in schema
  A_version_mismatch_throws_SchemaVersionMismatchException                  -> version mismatch
  A_missing_schema_version_is_reported_as_a_missing_required_field
  An_unreadable_enum_value_is_malformed_not_a_default
  A_null_where_the_model_requires_a_value_is_malformed_not_a_default
  The_four_error_types_are_distinct_and_share_one_base
  "An unknown required field" is covered from both sides deliberately: a required field the
  DOCUMENT does not declare, and a field the SCHEMA does not declare. Either one defaulting
  silently is the failure this line exists to prevent, so both are typed and both are tested.

DoD 4: GameState is a fully serializable tree: a test constructs a non-trivial state, serializes
       it, deserializes it, and asserts deep equality.
$ dotnet test IC2.sln --nologo -v q --filter "FullyQualifiedName~GameStateSerializationTests"
Passed!  - Failed: 0, Passed: 6, Skipped: 0, Total: 6 - IC2.Engine.Tests.dll (net10.0)
  A_non_trivial_state_round_trips_with_deep_equality
  The_fixture_state_is_actually_non_trivial        <- guards the fixture from being weakened:
      asserts an embarked army, a carrying fleet, a fleet under construction, a besieged city,
      a non-empty garrison, an occupied mercenary pool, a non-empty news log, a NEGATIVE
      diplomatic relation, and both a mercenary and a regular unit slot.
  Deep_equality_is_by_value_not_by_reference
  A_single_changed_leaf_anywhere_in_the_tree_breaks_equality
  A_changed_unit_slot_deep_inside_an_army_breaks_equality   <- proves the equality has teeth at
      the deepest level of the tree, not just at the root.
  A_save_round_trips_and_keeps_the_state_it_wraps

DoD 5: No gameplay constant is hardcoded in C#: a test asserts that every ruleset-governed number
       the model exposes is sourced from the loaded Ruleset object.
$ dotnet test IC2.sln --nologo -v q --filter "FullyQualifiedName~NoHardcodedConstantsTests"
Passed!  - Failed: 0, Passed: 5, Skipped: 0, Total: 5 - IC2.Engine.Tests.dll (net10.0)
  Every_number_the_ruleset_exposes_is_present_in_the_file_it_was_loaded_from
      RulesetNumbers.Enumerate walks the loaded Ruleset and yields (path, value) for every
      numeric leaf; each path is resolved in the file's own JSON and the values compared.
  Changing_every_number_in_the_file_changes_every_number_the_model_exposes
      Every number in the ruleset JSON is rewritten as v*2+7, the file is reloaded, and every
      exposed number must have moved with it. A value that came from a C# literal could not
      track the file and would fail here. This is the check that proves sourcing.
  The_ruleset_surface_covers_every_subsystem_later_tasks_consume
      18 named paths, one per subsystem, so a later removal is caught here.
  The_model_declares_no_numeric_constants_beyond_the_allowlist
      Reflection over every type in IC2.Engine.Model and IC2.Engine.Serialization: no const or
      static readonly numeric field is permitted except GameDataSchema.CurrentVersion, which is
      a file-contract identifier rather than a rule and is allowlisted by name in the test.
  Two_rulesets_with_different_numbers_drive_the_model_differently
      The fortification encoding decodes the same stored word differently under a different
      radix, which is only possible because FortificationCode reads the radix from the ruleset.

Supporting (task Scope line, not a numbered DoD): every field the task entry requires the model
to carry from day one has a named test.
$ dotnet test IC2.sln --nologo -v q --filter "FullyQualifiedName~ModelCoverageTests"
Passed!  - Failed: 0, Passed: 14, Skipped: 0, Total: 14 - IC2.Engine.Tests.dll (net10.0)
  purses+supply / army morale (+14) / regular-mercenary marker / fortification >100 encoding /
  fortify-then-siege-wipe / fleet condition + construction countdown / symmetric relation matrix
  with negative cooldowns / matrix sized to the world not to 16 / news ring buffer at ruleset
  capacity incl. overflow / terrain tile types as an OPEN list (a 13th type loads unchanged) /
  both terrain encodings agreeing / ruleset flags / mercenary pool slots / turn-order table.
```

## Provenance checklist

- [x] Every constant introduced or changed traces to a `tests/fixtures` entry or a cited report under `docs/reports/`.
- [x] Every `[designed]` value states **what was searched and came up empty** before it was designed (design-audit.md §4.5, promoted to a merge gate).
- [x] Nothing was invented to fill a gap: a value that could not be found in a report is omitted and listed below, not guessed.

`tests/fixtures/**` does not exist yet — that is T04, which runs in parallel with this task — so every
number here cites its report filename directly, in a `"_provenance"` entry sitting next to the value,
exactly as `game-design.md` §"Ruleset format" specifies. The reports live in the research repo
(`diegoami/imperial-conquest-2-research`, `docs/reports/`), per the repo split in `docs/HANDOVER.md`.

Reports the shipped constants are transcribed from:
`unit-type-stat-table-in-dat.md` (the five unit types' full stat table),
`decompiled-unit-map-orders-and-record-fields.md` (`+0x26` power weights, army `+14` morale, unit slot
`+0` marker, fleet `+20` condition, fortification's `>100` encoding, purses, supply capacities and
price, every naval and army-management cap, marker size bands),
`terrain-move-cost-table-in-dat.md` (the twelve-entry terrain table verbatim),
`decompiled-diplomacy-peace-terms-and-instant-battles.md` (relation states and cooldowns, thaw,
reparations, the instant resolver's arithmetic, naval combat, unity cap, 270 BC / 250 BC),
`decompiled-turn-and-calendar-sequencing.md` and `decompiled-quarterly-billing-and-economy.md`
(calendar and quarterly economy), `decompiled-recruitment-cost-formula.md`,
`fleet-order-at-caere.md`, `decompiled-news-log-identified.md` and `decompiled-sav-file-layout.md`
(40 slots x 61 bytes, 50 mercenary slots), `decompiled-defection-and-siege-attrition.md` and
`galatia-elimination-and-city-resupply-confirmed.md` (loyalty floors),
`battle-recording-melee-cap-confirmed.md`, `decompiled-combat-formula-structure.md` and
`combat-type-effectiveness-matrix.md` (the reserve detailed-resolver constants).

`[designed]` values, each with its search statement written into the file's own `_provenance`:
- `terrain.defaultMoveCost = 1` — `game-design.md` §Movement states the rule. Searched the reports for
  any original handling of an unpriced terrain code: `terrain-move-cost-table-in-dat.md` establishes the
  table covers exactly codes 0–11 and that codes ≥ 12 are markers, not terrain, so the original has no
  unpriced-terrain case to copy.
- `combat.scatteredDefeat.{survivorCasualtyNumerator, scatterTilesMin, scatterTilesMax}` = 40, 2, 4 —
  the numbers `game-design.md` §"The defeated side's fate" states. Searched the reports for any
  partial-defeat or retreat outcome: all three decompiled resolvers annihilate the loser
  unconditionally, so there is no original number to transcribe. Documented placeholders, to be retuned
  from play.
- The toy world's own fixture values (treasuries, populations, tribute, map layout) are labelled
  `designed: toy fixture value` in the file. They are not claims about the original: the shipped
  `classical-mediterranean` world with real figures is T29's deliverable.

Values omitted because no source was found (or "none"):

- `mapMarkers.cityPopulationTierThresholds` — **ships empty**. `game-design.md` §"City and army markers
  scale with size" records size-tiered city markers as a real requirement and states plainly that "the
  exact tier boundaries and how many tiers the original used are not yet decompiled". I searched the
  reports for a city-marker banding rule the way the army and fleet bands were found: the army and
  fleet bands are explicit functions (`FUN_0044A80C` / `FUN_0044A878`, in
  `decompiled-unit-map-orders-and-record-fields.md` part 3), but city marker codes 20–99 appear only as
  a range the movement walk refuses to step on, with no analogous banding function anywhere. Rather
  than invent thresholds, the field exists and is empty; T24 or a later decompilation pass fills it as
  data, with no model change.
- Unit-type table field `+0x20` (LI 18 · HI 2 · Ar 18 · LC 15 · HC 4) — **omitted**.
  `unit-type-stat-table-in-dat.md` lists it as unidentified and `docs/HANDOVER.md` still carries it as
  open. Transcribing an unnamed field into a named ruleset property would be inventing a meaning, so it
  is left out; when it is identified, it is a ruleset addition.
- Rebellion-check and weather-effect thresholds — **omitted**. `game-design.md` §Economy marks both
  `[open]` (`FUN_0044c204` / `FUN_004511bc` never decompiled). No placeholder is invented here; T08 owns
  the structure and will carry its own tagged values.

## Determinism

- [x] No `System.Random`, wall-clock reads, `Guid.NewGuid`, or order-dependent iteration in gameplay paths; every random draw goes through `IRng`; a seeded test proves reproducibility (or this PR touches no gameplay path — say which).

This PR introduces **no randomness at all** — `IRng` is T03's to define and is deliberately not
referenced here. Concretely:

- No `System.Random`, `DateTime.Now`/`UtcNow`, `Guid.NewGuid`, `Environment.TickCount` or
  `Stopwatch` anywhere in `src/IC2.Engine/Model` or `src/IC2.Engine/Serialization`, or in the tests.
  (A first draft of one test used `Guid.NewGuid()` for a temp directory name; it was replaced with a
  fixed name before commit, so nothing in this diff reaches for a generated identifier.)
- `SaveGame` deliberately carries **no timestamp**. Saves are named by `label` and ordered by the
  state's `calendar.turnIndex`.
- No `Dictionary`/`HashSet` is ever *enumerated* to produce state or output. Every collection in the
  model is `ValueList<T>`, which has exactly one stable order; the only hash containers in the diff are
  local duplicate-detection sets inside validation and a per-type contract cache, none of which is
  iterated. Id lookup is a linear scan over an ordered list.
- `GameDataRepository.Load` sorts filenames with `StringComparer.Ordinal` before reading, so the same
  directory always produces the same repository.
- Determinism is also what forced `ValueList<T>` and `ProvenanceMap` to exist: both preserve insertion
  order and have value equality, which is what makes "serialize → deserialize → serialize is identical"
  and "deep equality" checkable at all.

## Notes for the reviewer

**One evidence conflict found, carried rather than resolved.** `ruleset.siege` ships *two* fields for
what may be one adjustment: `defenderOwnerNotAllegiancePenaltyPercent: 20` (from
`decompiled-city-capture-resolution.md`, a defender penalty when owner ≠ allegiance) and
`attackerIsAllegianceDefenderReductionPercent: 10` (from `FUN_0044B27C`'s ×9/10 branch when the
*attacking* nation equals the allegiance). `design-audit.md` §2.13 flags these as unreconciled and notes
`FUN_0044A98C` was never decompiled. I did not pick one. Both are in the data with the conflict written
into their `_provenance`, so T17 (city capture / siege) inherits a visible decision rather than a silent
one. Flagging it here per §6.6 case 4.

**Two open items recorded in the data, not silently smoothed over.**
`economy.supplyTonsPerTalent`'s provenance carries `design-audit.md` Q9 (the code charges `amount/5` but
three video frames show army money unchanged) — the code's formula is what is transcribed.
`combat.detailedResolver.typeEffectiveness` carries `combat-type-effectiveness-matrix.md`'s own caveat
that the attacker/defender axis assignment is a reasoned inference; `typeEffectivenessOrder` records
which axis order the rows use, so a later correction is a data change.

**What DoD 5's check does and does not cover.** It proves that every number the `Ruleset` *exposes*
comes from the loaded file (by mutating the file and watching every exposed number move), and that no
numeric `const`/`static readonly` exists in the two namespaces. It cannot see a bare numeric literal
inside a method body; the structural defence against that is that the model has almost no behaviour —
the one place engine code could plausibly have embedded a rule is `FortificationCode`, and its methods
take the `CityOrderRule` from the ruleset rather than assuming 100, which
`Two_rulesets_with_different_numbers_drive_the_model_differently` proves. T03's source-scanning
determinism guard is the natural place for a broader literal sweep.

**Deliberate design choices a reviewer may want to weigh in on, none of which change a DoD line:**
1. *Sentinels modelled as absent values.* The original's `-1` (army not on the map) and `0xFFFF`
   (fleet launched) become `int?`/`string?` nulls, so no magic number appears in C#. The one raw
   sentinel kept verbatim is the unit slot's `mercenaryLabel`, because the value is a name-table index
   as well as a marker and has to survive a round-trip; `IsMercenary`/`IsRegular` wrap the comparison.
2. *Strict-by-default loading.* Validation runs over the parsed document against the record's reflected
   JSON contract *before* deserialization, so errors name a full path (`seats[1].control`) and never
   depend on `JsonException` message text.
3. *Two terrain encodings ship* (`runLength`, `base64`). The toy world uses run-length so it is readable
   by eye; base64 is there so T29's 44,800-cell export fits the same schema without widening it. Both
   are tested to agree.
4. *The toy ruleset is fully populated*, not minimal. Its numbers duplicate what T04's fixtures corpus
   will hold and what T29's `classical-faithful.json` will ship. That is intentional under the task's
   anti-widening hazard — the `Ruleset` record has to declare every group the later tasks read, and a
   required field needs a value in a loadable file — but it does mean two places will carry the same
   constants until T04 lands. If the reviewer prefers the toy ruleset to reference fixture ids once T04
   is merged, that is a clean follow-up.
5. `src/IC2.Engine/Placeholder.cs` (T01's) is left in place rather than deleted, since it is outside
   this task's Owns list.

**Not verified, stated honestly:** I did not run the original game or read any original DAT/SAV bytes —
every constant here is transcribed from the reports, which is the intended path (the repo split makes
`docs/reports/` read-only to this repository, and nothing from the research checkout was copied in).
The `_provenance` strings name report filenames rather than linking them, because those files are not in
this repository.


diegoami:
# Reviewer report — T02 (PR #33)

**Verdict: `status:rework`.** Gates 3 (determinism) and 4 (scope) pass cleanly. Gate 1 passes for
DoD 1/2/3/4 but fails on DoD 5. Gate 2 is strong — I re-derived ~45 constants from the research
reports and every one matched exactly — but has one finding. Gate 5 found four defects the suite does
not cover, one of which is also the concrete proof that DoD 5's guard has a hole.

This is **round 1**, not round 2 — no escalation needed. The fixes are all local and small; the
architecture, the provenance discipline and the test design are genuinely good and should not be
rebuilt.

Everything below was reproduced by me at `3a197c7` in the main checkout. Where I verified a finding by
running code, I did it in a scratch project **outside** the repository; the working tree is clean and
I changed nothing.

---

## Gate 1 — DoD, reproduced by me: **FAIL (DoD 5)**

Build and suite, run by me, not taken from the PR body:

| Command | Result |
| --- | --- |
| `dotnet build IC2.sln -warnaserror` | succeeded, **0 warnings**, 0 errors |
| `dotnet build IC2.sln -c Release -warnaserror` | succeeded, **0 warnings**, 0 errors |
| `dotnet test IC2.sln` | 45/45 engine, 1/1 data |
| `dotnet test -c Release --no-build` (what CI runs) | 45/45, 1/1 |

Per-line, each test class run on its own:

| DoD line | Check | Result |
| --- | --- | --- |
| 1 round-trip × 4 kinds | `RoundTripTests` | 5/5 pass |
| 2 toy scenario resolves world + ruleset | `ScenarioLoadingTests` | 5/5 pass |
| 3 three distinct typed errors | `TypedLoadErrorTests` | 9/9 pass |
| 4 `GameState` deep equality | `GameStateSerializationTests` | 6/6 pass |
| 5 no hardcoded gameplay constant | `NoHardcodedConstantsTests` | 5/5 pass |

Every DoD line has a real runnable check — none is asserted in prose only. DoD 1–4 are satisfied, and
several are satisfied more strictly than required (DoD 3 covers four error types, not three; DoD 4
asserts a changed leaf *breaks* equality, which is what makes the equality assertion mean something).

**DoD 5 is not satisfied.** You flagged the limit yourself — the guard bans numeric constant *fields*
and cannot see a literal in a method body. I judged whether that is a real weakening rather than a
theoretical one, and it is: the hole already hides a live defect.

`src/IC2.Engine/Model/GameStateFactory.cs:134`

```csharp
var calendar = new CalendarState(
    Week: ruleset.Calendar.WeekStep,   // <- the step, used as the starting week
```

`CalendarRules` has no `StartWeek` field, so the starting week is improvised in C# from an unrelated
field — which is precisely the class of thing DoD 5's first clause ("No gameplay constant is hardcoded
in C#") forbids, and `NoHardcodedConstantsTests` cannot see it because it is not a field. It is also
**wrong**, see gate 5 finding 1.

Two smaller instances of the same hole: `GameDataValidation.RequireUnitRange` hardcodes the 0–1
personality range (borderline — arguably a schema bound, not a tunable rule), and the test's own
docstring claims "No numeric constant exists anywhere in the model or serialization namespaces",
which is stronger than what `field.IsLiteral || (field.IsStatic && field.IsInitOnly)` actually checks
(it also misses instance `readonly` fields and expression-bodied properties).

Asked for, minimally: add `startWeek` (and, if you want the calendar fully data-driven,
`startSeasonIndex`) to `CalendarRules` with provenance, source `CalendarState.Week` from it, and
either narrow the test's docstring to what it checks or back the claim with a source scan. A source
scan is T03's determinism-guard pattern, so a narrowed docstring plus the new ruleset field is
sufficient here; please do not weaken the DoD line itself (§6.4).

## Gate 2 — Provenance: **FAIL (one finding)**

I re-derived a large sample directly from the local read-only copy of the research repo rather than
trusting the citations. **Everything I checked was exact**, including all six items I was asked to
scrutinise:

- **army morale `+14`** — `decompiled-unit-map-orders-and-record-fields.md:18-33`: `+14` is morale,
  seeds tactical morale, and multiplies both strength functions. `combat.powerTroopDivisor: 100`,
  `combat.powerDivisor: 80` and `siege.archerStrengthMultiplier: 3` reproduce
  `(Σ powerWeight[type] × troops / 100) / 80 × army[+14]` and the archers-tripled siege form exactly.
  `ArmyState.Morale`'s doc-comment correctly names §2.9's hazard, and I confirmed the **tactical
  morale array does not appear anywhere in this diff** — the model deliberately does not persist it.
  `newArmyMorale: 59` = `0x3B` ✓.
- **unit slot `+0`** — same report, mercenary-hire section: `0` = regular, non-zero = mercenary, value
  is the name-table index, and it selects the upkeep formula. `UnitSlot.MercenaryLabel` matches, and
  `ModelCoverageTests:63`'s `Gallic = 11` matches the report's three-save scan ✓.
- **fortification `>100`** — same report: `TFortifyCity_OK` writes `fortification += points × 100`;
  `FUN_0044B27C` starts `if (fort > 100) fort = fort % 100`; refused under siege and at 100; price is
  `cityRecord[+18] × points`. All five carried correctly ✓. (But see gate 5 finding 2 for the decoder.)
- **fleet condition `+20`** — same report:42-45. `launchConditionPercent: 100`, `launchSupplyTons: 50`,
  `repairCostDivisor: 5`, `constructionTicks: 24`, `combat.naval.conditionDivisor: 10` all exact ✓.
- **40-slot news ring buffer** — `decompiled-news-log-identified.md:5-13`: 40 slots, indices 0–39,
  61-byte (`0x3D`) slots, and `MostRecentSlot` as "index of the most recently used slot, which is why
  the save loops count+1 times". `NewsLog`'s shape and `-1`-for-empty match the report exactly ✓.
- **diplomatic cooldowns** — `decompiled-diplomacy-peace-terms-and-instant-battles.md`: `−8` broken
  trade, `−24` broken alliance, `−18` ended war, `−10` on the loser's surviving trade/alliance, `−8`
  for an ally still at war, thaw `+1` per quarter with a `1/3` chance of `min(0, v+3)`, the
  `while (sVar6 != 8)` 8-column bug, 3 trade partners, `W/4 + random(W/4) + cities × 10`. **All twelve
  values exact** ✓.

Also spot-checked and exact: the whole unit-type table (`unit-type-stat-table-in-dat.md` —
moves/battalion/shots/range/recruit/quarterly for all five types), the `+0x26` weights
20·100·40·60·120, the 12-entry terrain table at `0x1F622`, the 5×5 effectiveness matrix, every naval
instant-resolver constant, the loyalty floors 40/65/90, the caps (1,000 purse / 20 units / 100,000
troops / 198 armies / 50 merc slots / 990 unity / 30,000 / 40%), and 250 BC / 270 BC.

Two judgement calls I was asked about, both of which I agree with:

- **`siege`'s two unreconciled fields — correct, keep them.** `decompiled-city-capture-resolution.md:26`
  literally reads `if owner != allegiance: strength = strength × 4 / 5`, so `20` is right as read;
  `design-audit.md` §2.13's tail is the source for `FUN_0044B27C`'s `×9/10` branch and for the fact
  that `FUN_0044A98C` was never decompiled. The `_provenance` on both fields **genuinely documents the
  ambiguity** — it names both readings, says which report each came from, says they may be one
  adjustment or two, and explicitly hands the decision to T17. That is exactly what the plan's T17
  entry asks for. No escalation.
- **`typeEffectiveness` orientation** — the provenance says "confirmed values, candidate orientation"
  and attributes the attacker-row reading to the report's own reasoned inference. That is a faithful
  reading of `combat-type-effectiveness-matrix.md:19`'s caveat. Good.

### Finding 2.1 — `mapMarkers.cityPopulationTierThresholds`'s "searched and came up empty" is not accurate

`data/rulesets/toy-ruleset.json:414`. The provenance says the city marker codes appear in the reports
"**only** as a range the movement walk refuses to step on, with no decompiled banding function
analogous to `FUN_0044A80C`/`FUN_0044A878`".

The second half is true. The first half is not. `rivers-and-map-markers.md` ("City and unit overlays"):

> exactly **334 cells** have values from `20` through `199` … every `7.sav` city satisfies
> `map code = 20 + owner code + 16 × variant`, **with variants 0–4**. … The variant's meaning (**city
> size**, icon, or another display category) still needs confirmation.

and its own next-checks list item 3: *"Confirm the city-marker variant meanings."*

So the reports *do* carry evidence bearing directly on this: a **5-band** city marker encoding,
holding for 334/334 cities, whose leading candidate meaning is city size. That is the direct analogue
of the three army bands and three fleet bands, just without the thresholds attached. T11 and T24 need
the *tier count* at least as much as the thresholds.

This is the design-audit §2.7 pattern pointing the other way — a search reported as empty that was not
empty — and §4.5 (promoted to this gate by §6.2) is specifically about the accuracy of that
statement, so it has to be a finding even though the shipped value may well stay `[]`.

Asked for: keep the empty list if you still judge the thresholds unevidenced, but rewrite the
provenance to cite `rivers-and-map-markers.md`, state that the **count** (5 variants, `20 + owner +
16 × variant`, 334/334) is evidenced while the **population boundaries** are not, and say that the
report's own next-check is what would close it. If you conclude the 5 bands are strong enough to ship
as a tier *count* with unset thresholds, say that instead — either is defensible; the current text is
not.

Everything else on this gate passes.

## Gate 3 — Determinism: **PASS**

- No `System.Random`, `DateTime.Now/UtcNow`, `DateTimeOffset`, `Guid.NewGuid`, `Environment.TickCount`
  or `Stopwatch` anywhere in `src/IC2.Engine` or the new tests. `SaveGame`'s doc-comment calling out
  why there is no timestamp is the right instinct.
- No order-dependent dictionary enumeration. The three hash containers in the diff are all
  non-enumerated: `JsonContract.Cache` (a `ConcurrentDictionary` keyed by `Type`, lookup only) and the
  `HashSet`s in `ValueList.HasDistinctIds`/`FirstDuplicateId` and `GameDataValidation` (membership
  only). `ValueList<T>`'s remarks explaining that the model simply never stores a dictionary is the
  right structural answer, not a workaround.
- Serialization ordering is deterministic end to end: declaration-order members, `ProvenanceMap`
  preserving insertion order through its converter, `ValueList<T>` as an ordered array, one shared
  `JsonSerializerOptions`, and `GameDataRepository.LoadAll` sorting filenames with
  `StringComparer.Ordinal`. `RoundTripTests` proves byte-identity for all four kinds. This is the part
  of the diff every later round-trip test depends on and it is right.
- Even the test-side temp directory is a fixed name, with the reason cited. Noted and appreciated.

Nit (not blocking): `JsonContract.Build` at `src/IC2.Engine/Model/JsonContract.cs:111` picks the
constructor via `constructors.OrderByDescending(c => c.GetParameters().Length).FirstOrDefault()`.
`Type.GetConstructors()` has no documented order, so two public constructors with equal arity would
resolve by reflection order. Harmless today (every model record has exactly one public constructor),
but it is a latent order dependence in the one place the whole schema is derived from — worth a
deterministic tie-break or a comment.

## Gate 4 — Scope: **PASS**

All 30 changed files are inside the declared Owns list; `git diff --name-status main...HEAD` is
30 × `A`, 0 modifications, 0 deletions.

- `src/IC2.Engine/Model/**` — 12 files ✓
- `src/IC2.Engine/Serialization/**` — 6 files ✓
- `tests/IC2.Engine.Tests/Model/**` — 8 files ✓
- exactly the three named data files: `data/worlds/toy-3city.json`,
  `data/rulesets/toy-ruleset.json`, `data/scenarios/toy-3city.json` — and **nothing else** under
  `data/worlds/` or `data/rulesets/`, so T29's files are unobstructed ✓

Confirmed untouched: `docs/build-orchestration-plan.md` (and all of `docs/`), `IC2.sln`,
`Directory.Build.props`, `.github/workflows/`, `src/IC2.Data`, `src/IC2.Inspect`, `src/IC2.Cli`,
`godot/`, and `tests/IC2.Engine.Tests/IC2.Engine.Tests.csproj`.

**On not editing T01's `.csproj`: your call was right, and the mechanism is CI-safe.** I checked.
`TestPaths.FindRepositoryRoot` walks up from `AppContext.BaseDirectory` to `IC2.sln`; under
`.github/workflows/ci.yml` the assembly lands at `tests/IC2.Engine.Tests/bin/Release/net10.0/`, which
is inside the checkout, so the walk terminates at the repo root on the Linux runner exactly as it does
here. I ran the Release/`--no-build` path CI uses and it passes. Reading the committed files in place
rather than a copy is also strictly better for this task, since DoD 1 and DoD 5 are claims about the
*shipped* files. Keep it.

## Gate 5 — Correctness: **FAIL (4 findings)**

`/code-review --effort high` ran and is folded in below; I reproduced each of these myself rather than
relaying them.

### 5.1 (high) `GameStateFactory.cs:134` — the initial week's parity makes the season unable to advance

`CreateInitial` seeds `CalendarState.Week` from `ruleset.Calendar.WeekStep`, i.e. **2**.
`decompiled-turn-and-calendar-sequencing.md` confirms the cycle at code level:

```text
week = (week + 2) mod 12
if week == 1:            // wrapped from 11 back to 1
    season = (season + 1) mod 4
    if season == 0: year -= 1
```

Only an **odd** starting week ever reaches 1. Starting at 2 gives 2 → 4 → 6 → 8 → 10 → 0 → 2 → … so
the season never advances and the BC year never decrements. Every save the project has (Week 9,
Week 11 Spring 270 BC) is on the odd parity.

Verified by running the shipped data through the built engine:

```
initial week = 2  (weekStep=2, modulus=12, seasonAdvanceFromWeek=11)
weeks: 4,6,8,10,0,2,4,6,8,10,0,2,4,6
ever reaches week 1 (season advance): False
```

No test catches this because `ToyFixtures.NonTrivialState` overwrites `Calendar` with a hand-written
`CalendarState(Week: 7, …)`, and nothing asserts the *initial* calendar. T06's calendar DoD is a
byte-equal week/season/year sequence, so it would inherit a start state from which its own DoD cannot
pass.

Fix: a real `calendar.startWeek` (1) in `CalendarRules` + the toy ruleset with provenance, read here;
and an assertion in `ScenarioLoadingTests` on the initial `CalendarState`. This is the same change
that closes the DoD 5 gap above.

### 5.2 (medium) `FortificationCode.cs:44` — `MaxOrderablePoints` ignores points already pending

It returns `MaxPercent − FinishedPercent`, dropping `PendingPoints`. Verified against the shipped
ruleset:

```
code=340  finished=40  pending=3  maxOrderable=60
after ordering 60 -> code=6340  finished=40  pending=63  = 103%  (maxPercent=100)
```

`ModelCoverageTests.cs:100` asserts `Assert.Equal(60, …)`, so the test currently locks the behaviour
in. The original forbids the situation a different way — `TFortifyCity` refuses outright with *"This
city is already being fortified"* when `fort > 100` — so the faithful answer is probably that
`MaxOrderablePoints` returns `0` while an order is pending (or the helper gains a `CanOrder`
companion). Either way, a helper that hands a caller a number which overshoots `MaxPercent` is a trap
for T-city-orders.

Related, lower: `IsOrderInProgress` at line 19 compares against `InProgressEncodingRadix` while
`MaxOrderablePoints` compares against `MaxPercent`. They agree only because the shipped ruleset sets
both to 100. With `MaxPercent: 100, InProgressEncodingRadix: 1000`, `WithOrder(0, 1)` = 1000 decodes
as "finished 1000%, nothing pending". Since rulesets are data by design, please make the threshold
`MaxPercent` and keep the radix for the arithmetic only.

(Separately: at the shipped radix, a 0% city ordering exactly 1 point writes `100`, which reads back
as "fortified 100%, nothing pending". That one **is** faithful — the original's `fort > 100` has the
identical ambiguity — so I am not asking you to change it, only to add a one-line comment saying it is
deliberate, because the next reader will file it as a bug.)

### 5.3 (medium) `GameDataValidation.cs:197` — `ValidateState` never resolves nation references

`ValidateWorld` calls `RequireNation` for every city owner/allegiance and every army/fleet nation.
`ValidateState` checks only `TurnOrder` and `Relations.NationIds` — not `CityState.Owner`,
`CityState.Allegiance`, `ArmyState.Nation` or `FleetState.Nation`. Verified:

```
save with cities[0].owner = "atlantis"  ->  LOADED OK; NationById("atlantis") -> null
```

Given this file's own premise ("Catching them here means no later task has to defend against…"), and
given that every later task will write `state.NationById(city.Owner)!`, this is the gap most likely to
turn into a downstream NRE. Same `RequireNation` treatment, against `state.Nations`.

### 5.4 (medium) `World.cs:110` — the run-length guard overflows, and the failure escapes untyped

`if (written + run.Count > expected)` is unchecked `int` arithmetic, so a run of `count: 2147483647`
wraps the sum negative, passes the guard, and runs the fill loop off the end of `cells`.
`ValidateWorld` catches only `InvalidOperationException`, so the result leaves the loader untyped:

```
runs = [{code:2,count:1},{code:2,count:2147483647}]  ->  UNTYPED ESCAPE: IndexOutOfRangeException
```

That is a malformed file producing an untyped exception, which is the contract DoD 3 exists to
guarantee. `checked(width * height)` on line 86 has the same escape (`OverflowException` past the same
catch). Compare as `run.Count > expected - written`, and widen the catch in `ValidateWorld`.

### Lower-severity, non-blocking

- `GameState.cs:375` — `NewsLog.Append` decides fullness from `MostRecentSlot`, but
  `GameDataValidation` only requires `MostRecentSlot ∈ [-1, Slots.Count-1]`, never that the two agree.
  A loaded log with 40 slots and `mostRecentSlot: 5` takes the not-full branch and grows past
  `ringBufferSlots`. Gate the capacity test on `Slots.Count`, or have the validator require
  `MostRecentSlot == Slots.Count - 1`.
- `GameDataLoader.cs:28` — `LoadFile` catches only `IOException`; `File.ReadAllText` also throws
  `UnauthorizedAccessException` (path is a directory, or ACL-denied), which escapes untyped and breaks
  the `catch (GameDataException)` contract `GameDataRepository.Load` documents.
- `GameDataLoader.cs:108` — the schema version is checked on the root only, so a `SaveGame`'s nested
  `state.schemaVersion` is unchecked. Verified: `state.schemaVersion = 99` loads fine.
  `ValidateSave` already cross-checks the state's world/ruleset/scenario ids; the version belongs
  beside them.
- `GameDataValidation.cs:245` — the army↔fleet embark link is validated one way only. Nothing requires
  `army.AboardFleetId` and `fleet.CarriedArmyId` to be mutual or exclusive, so two armies may both
  claim the same fleet that `FleetState` documents as carrying "the one army aboard".
- `GameStateFactory.cs:16` — the `<exception>` doc promises a duplicate-id check the method does not
  perform.
- `ProvenanceMap.GetHashCode` does not mix in the entry count (`ValueList` does). Cosmetic.

### `ValueList<T>` and `ProvenanceMap`, judged specifically

You asked for these to be checked rather than accepted, since every later round-trip test rests on
them. **Both are correct.** Equality is ordered, length-guarded and element-wise via
`EqualityComparer<T>.Default` (so null elements are handled); `GetHashCode` uses `HashCode` and is
consistent with `Equals`; `Equals(object)` handles null and the wrong type; the type is genuinely
immutable (both the `IEnumerable<T>` constructor and `Of` copy, so no aliasing through a caller's
array); `Empty` is safe to share; the converter emits a plain JSON array and rejects `null`.
`ProvenanceMap` likewise preserves order exactly, compares ordered, and its converter round-trips a
flat string object faithfully and rejects non-object/non-string shapes. `GameStateSerializationTests`
pins the behaviour that matters (value equality across two independent constructions, equal hash
codes, and a changed leaf at two depths breaking equality). I have no changes to ask for here.

### Model-completeness notes for later tasks (not findings)

Not blocking — T02's enumerated must-carry list is fully satisfied — but flagging while it is cheap,
since the point of this task is that nobody has to widen the model later:

- `SiegeRules` carries the archer multiplier and `/80`, but not the confirmed defender-strength
  weights from `decompiled-city-capture-resolution.md` (`fortification × 150 + loyalty × 250 +
  <third> × 200`, the `× 5/3` branch above fortification 59, and `+ garrison/2`). T17 will need them.
- `NavalCombatRules` has no unity-swing divisor for the confirmed `± floor(loserShips / 2)`.
- `UnitTypeRules` omits the table's `+0x20` (18/2/18/15/4), still unidentified in the reports —
  reasonable to omit, worth a one-line note in `_provenance` saying it was deliberately dropped.

---

## Second opinion (`/code-review --effort ultra`) — plain statement of what ran

Per §3.4/§6.3, T02 gets an additional opinion. I invoked `/code-review --effort ultra` on this PR and
it **did** produce a review, which I have folded in above — but I must report plainly that it ran as a
local forked review in my session, **not** as the cloud multi-agent ultra path, and it could not post
to the PR from the user's account non-interactively. So treat it as a second independent pass over the
same diff rather than as the full ultrareview the plan describes; if the orchestrator wants the real
cloud pass, it needs an interactive session.

For the record, it independently reached 5.2 (the `MaxOrderablePoints` double-order, which I had not
found on my own), 5.3, 5.4 and the news-log and schema-version items, and the `--effort high` pass
independently reached 5.1. Neither pass disagreed with the other, nor with me, on anything.

---

## Summary

| Gate | Result |
| --- | --- |
| 1 DoD reproduced | **FAIL** — DoD 1–4 pass; DoD 5's guard has a hole that hides a live defect |
| 2 Provenance | **FAIL** — one finding (2.1); ~45 constants re-derived and exact, siege conflict handled correctly |
| 3 Determinism | **PASS** |
| 4 Scope | **PASS** |
| 5 Correctness | **FAIL** — 5.1 high, 5.2/5.3/5.4 medium, plus lower-severity items |

Must-fix before re-review: **5.1**, **5.2**, **5.3**, **5.4**, **2.1**, and the DoD 5 gap (which 5.1's
fix largely closes). The lower-severity list is worth doing while you are in the file but I will not
block on it.

Nothing here needs a human decision. The one genuine evidence conflict in the diff — siege's `20%`
vs `10%` — is correctly carried as two named flags with the ambiguity written down, which is exactly
what the plan's T17 entry prescribes, so it does **not** trip §6.6 case 4.

diegoami:
# Rework round 1 — all six must-fix findings addressed (`7b96aee`)

Thank you for the review — it was specific enough to act on directly, and 5.1 in particular was a real
defect that my own DoD 5 test was structurally unable to see. Every must-fix is addressed below, plus
all six lower-severity notes. Nothing was rebuilt: the architecture, `ValueList<T>`/`ProvenanceMap`,
the contract-driven validation and the mutation test are unchanged, as you asked.

**Gates: build 0 warnings, 59/59 engine tests (was 45), 1/1 data tests.** 14 new tests, all of them
negative or invariant tests attached to a specific finding.

---

## 5.1 (high) + the DoD 5 failure — the starting calendar

You are right, and the parity argument is the important half. Fixed at the root rather than at the
call site:

- `CalendarRules` gains **`startWeek`** and **`startSeasonIndex`** as real ruleset fields.
  `GameStateFactory` reads all four calendar values from the ruleset; nothing is improvised.
- `startWeek: 1`, cited to `impconq2-full-save-analysis.md` (`Week 1      Spring      270 BC`, the
  earliest dated news slot) and `one-turn-save-comparison.md` (the following turn reads `Week  3`).
  `startSeasonIndex: 0`, cited to `saves-and-screenshots.md` / `strategic-recording-and-summer-saves.md`
  (trailer `+44` is `0` in all five Spring saves, `1` in all three Summer saves). The `_provenance`
  spells out the parity reasoning and notes that every dated save in the project (weeks 1, 3, 5, 7, 9,
  11) is on the odd parity.
- **`GameDataValidation.ValidateCalendar`** now rejects, at load, any calendar whose start week cannot
  reach `seasonAdvanceFromWeek` by stepping `weekStep` — so this defect class cannot return through
  data either. It also range-checks the four cycle fields.
- **`ToyFixtures` no longer masks the start.** It derives its mid-game calendar by advancing the
  initial one through the confirmed cycle (`AdvanceWeeks`, a clearly-labelled *test* helper kept out of
  `src/IC2.Engine` so it cannot be mistaken for T06's implementation). That was the thing that hid the
  bug, so removing the mask matters as much as the fix.

Three new tests pin it: `The_initial_calendar_comes_entirely_from_the_loaded_ruleset` (including an
explicit `Assert.NotEqual(WeekStep, Week)` — the exact confusion, asserted against),
`The_initial_week_can_actually_reach_the_season_boundary` (walks the cycle and names every week it
visited on failure), and `Advancing_the_shipped_start_advances_the_season_and_then_the_year` (6 turns →
season 1, 24 turns → year 269). Plus
`A_calendar_whose_start_week_can_never_reach_the_season_boundary_does_not_load` and
`The_calendars_starting_week_is_ruleset_data_not_an_engine_value` in the DoD 5 class.

On the two smaller instances you named: the DoD 5 docstring is **narrowed** to state exactly what the
reflection check sees (`const` and `static readonly` numeric *fields*, not instance readonly fields,
not expression-bodied properties, not method-body literals), with a paragraph saying that checks 1–2
are what actually cover the values and that a source-text sweep belongs with T03's guard. I did **not**
add a source scan — duplicating T03's guard here seemed worse than pointing at it, but say the word if
you'd rather have it. `RequireUnitRange`'s 0–1 bound stays in code with a comment explaining why it is
a schema bound rather than a tunable rule: `AiPersonality`'s parameters are *defined* as fractions in
`game-design.md` §AI, so a value outside that range is a malformed document, not an unusual ruleset.

## 5.2 (medium) — `MaxOrderablePoints`

Fixed as you suggested, and the test that locked in the bug is updated:

- `MaxOrderablePoints` returns **0 while an order is pending**, matching the original's outright refusal
  (*"This city is already being fortified."*). A new `CanPlaceOrder` companion carries both refusal
  conditions. The `<remarks>` says why it does not return `max − finished − pending`.
- **`IsOrderInProgress` now tests `MaxPercent`, not `InProgressEncodingRadix`.** New test
  `The_in_progress_threshold_is_the_maximum_not_the_encoding_radix` uses your exact example
  (`MaxPercent: 100, InProgressEncodingRadix: 1000`) and asserts `WithOrder(40, 3)` = 3040 decodes as
  40% finished with 3 pending.
- The 0%-plus-one-point ambiguity is left alone with the one-line comment you asked for, saying it is
  deliberately faithful to the original's identical `fort > 100` ambiguity.

`ModelCoverageTests` now also asserts that a siege wipe restores orderability (`340` → `40` → 60
points available), which is the behaviour the fix is meant to preserve.

## 5.3 (medium) — `ValidateState` nation references

`ValidateState` now runs `RequireStateNation` over every `CityState.Owner`, `CityState.Allegiance`,
`ArmyState.Nation` and `FleetState.Nation`, against `state.Nations`. I also closed the one-way embark
link you flagged in the lower-severity list at the same time: the army↔fleet reference must now be
**mutual**, so two armies cannot both claim one fleet.

Three new negative tests: `A_state_naming_a_nation_that_does_not_exist_does_not_load`,
`An_army_whose_nation_does_not_exist_does_not_load`, `A_one_sided_embark_link_does_not_load`.

## 5.4 (medium) — the run-length overflow

`written + run.Count > expected` → `run.Count > expected - written`. The map-size computation is now a
`long` with an explicit range check rather than `checked(...)`, so an absurd size is an
`InvalidOperationException` the loader turns into a typed error instead of an `OverflowException`
escaping past `ValidateWorld`'s catch. The base64 length comparison is widened to `long` for the same
reason.

`An_overlong_terrain_run_is_malformed_not_an_untyped_runtime_exception` uses your reproduction
(`count: int.MaxValue` after a 1-cell run).

## 2.1 (gate 2) — the city-marker provenance

You are right and the correction is the one that matters most to me, because it is exactly the
`design-audit.md` §2.7 pattern I was supposed to be guarding against. I had searched for a *banding
function* and reported the absence of one as an absence of evidence, which is not the same thing.

The list still ships **empty**, but the `_provenance` is rewritten to state both halves precisely: that
`rivers-and-map-markers.md` establishes a five-value city-marker variant axis
(`map code = 20 + owner code + 16 × variant`, variants 0–4, holding 334/334), that the same sentence
says the variant's meaning — "city size, icon, or another display category" — still needs confirmation,
that the report's own next-check item 3 is what would close it, and that no population boundary is
recorded anywhere. The note also opens by saying an earlier revision claimed the search came up empty
when it had not.

**On your "either is defensible" offer, I chose the conservative side deliberately**: shipping 5 as a
*size-tier count* would assert the half the report explicitly leaves unconfirmed (that the variant means
size rather than icon shape). I would rather T24 read a note that says "five display variants exist,
meaning unconfirmed" than a ruleset field that says "there are five size tiers". Happy to flip it if you
disagree — it is one line plus a field.

---

## Lower-severity notes — all six folded in

| Note | What changed |
| --- | --- |
| `NewsLog.Append` fullness from `MostRecentSlot` | Fullness is now decided from `Slots.Count` and the log can never grow past capacity; `IsConsistent()` added and `ValidateState` requires `MostRecentSlot == Slots.Count - 1`. Test: `A_news_log_whose_index_disagrees_with_its_slots_does_not_load`. |
| `LoadFile` catches only `IOException` | Now also `UnauthorizedAccessException` and `NotSupportedException`, with the reason in a comment. Test: `Reading_a_directory_instead_of_a_file_is_a_typed_error`. |
| Nested `state.schemaVersion` unchecked | `ValidateSave` checks it beside the ids, throwing `SchemaVersionMismatchException`. Test: `A_saves_nested_state_version_is_checked_too`. |
| One-way embark link | Now mutual — see 5.3. |
| `GameStateFactory`'s `<exception>` doc | Rewritten to match what the method actually checks, and to say duplicate ids are `GameDataValidation`'s job. |
| `ProvenanceMap.GetHashCode` | Mixes in the entry count, matching `ValueList`. |
| Gate 3 nit: `JsonContract` constructor order | Deterministic tie-break added (arity desc, then ordinal parameter-type signature), with a comment on why this one place deserves it. |

## Two anti-widening additions you flagged as cheap

Both taken, since preventing later widening is this task's whole point:

- **`SiegeRules`** now carries the confirmed defender-strength terms from
  `decompiled-city-capture-resolution.md`: `defenderFortificationWeight: 150`,
  `defenderLoyaltyWeight: 250`, `defenderUnidentifiedFieldWeight: 200`,
  `highFortificationThreshold: 59` with `highFortificationBonusNumerator/Denominator: 5/3`, and
  `defenderGarrisonTroopDivisor: 2`. Two honesty notes are in the provenance: the third term's *field*
  is unidentified in the report (so the property name says so rather than guessing), and the `×5/3`
  branch carries a further condition the report did not recover.
- **`NavalCombatRules.unitySwingShipDivisor: 2`** for the confirmed `± floor(loserShips / 2)`.
- **`+0x20`** is still deliberately omitted, now with the one-line note you asked for on the
  `unitTypes` provenance saying it was dropped because the field is unidentified, not overlooked.

---

## Re-run DoD evidence

```text
$ dotnet build IC2.sln --nologo -c Debug /t:Rebuild
Build succeeded.  0 Warning(s)  0 Error(s)

$ dotnet test IC2.sln --nologo -v q
Passed!  - Failed: 0, Passed:  1, Skipped: 0, Total:  1 - IC2.Data.Tests.dll (net10.0)
Passed!  - Failed: 0, Passed: 59, Skipped: 0, Total: 59 - IC2.Engine.Tests.dll (net10.0)

DoD 1  RoundTripTests                 5/5   (unchanged)
DoD 2  ScenarioLoadingTests           8/8   (was 5 — +3 initial-calendar tests, 5.1)
DoD 3  TypedLoadErrorTests           18/18  (was 9 — +9, covering 5.3, 5.4 and four
                                             lower-severity items; every one asserts a
                                             TYPED error, which is the contract 5.4 breached)
DoD 4  GameStateSerializationTests     6/6   (unchanged)
DoD 5  NoHardcodedConstantsTests       6/6   (was 5 — + startWeek/startSeasonIndex/startYearBc
                                             are ruleset-sourced; docstring narrowed)
       ModelCoverageTests            15/15  (was 14 — + radix-vs-maxPercent; the
                                             MaxOrderablePoints assertion is corrected)
```

**The two new tests were checked for teeth, not just for passing.** I temporarily reverted the 5.4
guard and the 5.3 reference check and re-ran only those two tests:

```text
$ dotnet test IC2.sln --filter "An_overlong_terrain_run_is_malformed|A_state_naming_a_nation_that_does_not_exist"
  TypedLoadErrorTests.An_overlong_terrain_run_is_malformed_not_an_untyped_runtime_exception [FAIL]
  TypedLoadErrorTests.A_state_naming_a_nation_that_does_not_exist_does_not_load             [FAIL]
Failed!  - Failed: 2, Passed: 0, Skipped: 0, Total: 2
```

Both fixes were then restored and the full suite re-run clean (the 59/59 above is after the restore;
`git diff` confirms neither revert survived).

## Scope

Still 30 files, all inside the Owns list — this round modified 15 of them and added none. Nothing under
`docs/`, no `IC2.sln`, no `Directory.Build.props`, no `.github/`, no `src/IC2.Data`/`IC2.Inspect`/`IC2.Cli`,
no `godot/`, and `tests/IC2.Engine.Tests/IC2.Engine.Tests.csproj` is still untouched. No new NuGet
package. No "Done when" line edited.

## One thing I did not change, and why

The `ruleset.siege` 20%-vs-10% pair stays as two named fields with the ambiguity in their provenance,
per your gate-2 finding. The new defender-strength weights sit alongside them, so T17 now inherits the
complete confirmed expression *and* the one unresolved question about it in the same place.

diegoami:
# Reviewer report — T02 (PR #33), re-review of rework round 1

**Verdict: all five gates pass. `status:approved`.** Every round-1 finding is fixed, and fixed at the
root rather than papered over. I re-derived the new constants and the four new citations from the
reports myself, verified each fix with my own probes rather than by watching the new tests go green,
and ruled on both judgement calls below. **This is not a round-2 disagreement** — there is nothing
here for the orchestrator to escalate.

Re-checked at `7b96aee`, working tree clean, nothing committed or pushed by me.

---

## Gate 1 — DoD, reproduced by me: **PASS**

| Command | Result |
| --- | --- |
| `dotnet build IC2.sln -c Debug -t:Rebuild -warnaserror` | succeeded, **0 warnings**, 0 errors |
| `dotnet build IC2.sln -c Release -warnaserror` | succeeded, **0 warnings**, 0 errors |
| `dotnet test IC2.sln` | **59/59** engine, 1/1 data |
| `dotnet test -c Release --no-build` (CI's path) | 59/59, 1/1 |

Per-class, each run on its own — every claimed count is exact:

| DoD line | Class | Round 1 | Now |
| --- | --- | ---: | ---: |
| 1 round-trip | `RoundTripTests` | 5/5 | 5/5 |
| 2 scenario resolves | `ScenarioLoadingTests` | 5/5 | **8/8** |
| 3 typed errors | `TypedLoadErrorTests` | 9/9 | **18/18** |
| 4 `GameState` deep equality | `GameStateSerializationTests` | 6/6 | 6/6 |
| 5 no hardcoded constant | `NoHardcodedConstantsTests` | 5/5 | **6/6** |
| scope coverage | `ModelCoverageTests` | 14/14 | **15/15** |

### Ruling on judgement call 1 — DoD 5 is now genuinely satisfied, and the narrowed docstring is not a weakening

I called DoD 5 failed last round, so I owe an explicit ruling. Taking §6.4's list of what an agent may
not do — edit the DoD, weaken an assertion to a range, mark a test `Skip`, delete a failing assertion —
**none of them happened**:

- The plan's DoD text is untouched (gate 4 confirms no `docs/` diff at all).
- The reflection check's **code is byte-for-byte the same strictness** as round 1. Nothing was
  loosened. What changed is the *name* and the *docstring*, which previously claimed "No numeric
  constant exists anywhere in the model or serialization namespaces" while the code only ever saw
  `const` and `static readonly` numeric **fields**.
- Correcting a comment that claimed more than its code checked is the *opposite* of weakening — it is
  the same over-claim pattern design-audit §2.7 exists to catch, applied to a docstring instead of a
  citation.
- Most importantly, the actual defect that over-claim was hiding is **gone at the root**, not excused:
  `startWeek` and `startSeasonIndex` are now real `CalendarRules` fields, so they fall inside checks 1
  and 2 — the mutation test that proves each exposed number tracks the file it was loaded from — which
  is where DoD 5's real enforcement lives.

I did not take the structural argument ("the model has almost no behaviour") on trust. I swept
`src/IC2.Engine/Model/*.cs` and `src/IC2.Engine/Serialization/*.cs` for numeric literals in
expressions myself. **The only hits are three in `TerrainGrid.Decode`'s base64 branch** — `* 2`,
`/ 2`, `<< 8` — which are the 16-bit little-endian cell width that `TerrainEncoding.Base64`'s own
doc-comment declares. That is a file-format constant of exactly the same kind as `schemaVersion`, not
a gameplay rule. No gameplay number is embedded as a bare literal anywhere in either namespace.

I also agree with declining to build a source-text sweep here. T03's Owns includes a determinism guard
that already scans `src/IC2.Engine` sources; a second scanner living in T02's test project would be
duplicated infrastructure T03 then has to reconcile with its own. Pointing at it is the right call, and
`RequireUnitRange`'s 0–1 bound is now correctly reasoned as a schema bound on a type
`game-design.md` §AI *defines* as fractions, rather than a tunable rule.

## Gate 2 — Provenance: **PASS**

### The four new calendar citations — all verified verbatim, no §2.7 repeat

This was the one that mattered most, since `startWeek` now defines the start state every later task
inherits. I opened each cited report:

- `impconq2-full-save-analysis.md` — *"Visible text includes `Week 1      Spring      270 BC`"* ✓
- `one-turn-save-comparison.md` — one sentence carries **both** quoted strings: the later save appends
  a slot reading `Week  3      Spring      270BC`, and *"The first save's last visible dated slot says
  `Week 1      Spring      270 BC`"* ✓
- `saves-and-screenshots.md` — *"Across all saves, `+44` is 0 for Spring and 1 for Summer, strongly
  supporting a season index."* ✓ Its trailer table also independently corroborates the odd parity:
  every dated label is 1, 3, 5, 7, 9, 11 ✓
- `strategic-recording-and-summer-saves.md` — *"Trailer `+44` is `0` in all five Spring saves and `1`
  in all three Summer saves"* ✓ — literally "all five" and "all three", exactly as the provenance
  paraphrases it.

Each citation contains the claim it is cited for. Good.

### The new siege and naval constants — re-derived as new scope, not accepted as fixes

You were right to flag these as new surface. `decompiled-city-capture-resolution.md`'s
`FUN_0044a98c` block reads:

```text
strength = fortification × 150 + loyalty × 250 + <third field> × 200
if <condition> and fortification > 59: strength = strength × 5 / 3
if owner != allegiance: strength = strength × 4 / 5
strength += (garrison troops assigned to this city) / 2
```

All seven new `SiegeRules` values are exact: 150, 250, 200, 59, 5, 3, 2 ✓. And **both honesty notes
are accurate, not decorative**:

- the report genuinely writes `<third field>` without naming which city field it reads, so
  `DefenderUnidentifiedFieldWeight` is the honest name — it carries the weight and refuses to guess
  the operand, and the provenance says T17 must identify it before the term can be implemented;
- the branch genuinely reads `if <condition> and fortification > 59`, with `<condition>` unrecovered,
  and `highFortificationThreshold`'s provenance says exactly that.

Naming a property after what is *not* known is the right instinct and I want it noted approvingly —
it is the structural version of the tagging discipline design-audit asks for.

`combat.naval.unitySwingShipDivisor: 2` matches
`decompiled-diplomacy-peace-terms-and-instant-battles.md`'s
`unity[loserNation] -= floor(loserShips / 2)` / `min(990, … + floor(loserShips / 2))` ✓.

The new root note explaining that unit-type field `+0x20` (18/2/18/15/4) is **deliberately** not
carried, because `unit-type-stat-table-in-dat.md` lists it as unidentified and naming a property for
it would invent a meaning, is a good catch — that was a nit from my last pass and it is answered better
than I asked for.

### Ruling on judgement call 2 — the empty tier list plus the rewritten note is the right resolution; do not flip it

The rewritten `cityPopulationTierThresholds` note satisfies design-audit §4.5, and exceeds its minimum.
§4.5 asks only that the document say what was searched and came up empty. This note goes further and
splits the question in two — what the reports **do** establish (`20 + owner + 16 × variant`, variants
0–4, holding 334/334) and what they **do not** (that same sentence leaves the variant's meaning
unconfirmed; no banding function; the report's own next-check item 3 is what would close it). I
re-checked `rivers-and-map-markers.md` against every part of that, **including that the next-check is
genuinely item 3** — it is. It also says plainly that an earlier revision over-claimed, which is the
right way to leave a corrected record.

On the substance: **I agree with the conservative side and do not want it flipped.** Shipping `5` as a
size-tier count would assert the half the report explicitly leaves open — that the variant means size
rather than "icon, or another display category". Leaving the requirement standing in
`game-design.md` §"City and army markers scale with size" while the ruleset stays silent is the
correct shape, and T11/T24 can choose a tier count as a *rendering* decision with its own provenance,
which is a cleaner home for it than a ruleset constant carrying an unconfirmed RE claim.

**One non-blocking nit, for T29 rather than for this PR.** `startSeasonIndex` is tagged `confirmed:`,
but both cited reports call `+44` a *"strong candidate"* / *"candidate season byte"*, and
`strategic-recording-and-summer-saves.md`'s own next-check item 2 is to verify it with a
Summer→Autumn sample. The underlying fact — the game starts in Spring — **is** directly confirmed by
the news text; only the Spring→0 *mapping* is candidate-strength. The provenance quotes the raw
observation so no reader is misled, and the value is right either way, so this does not affect the
gate. Worth matching the reports' own wording when T29 ships the real `classical-faithful` ruleset.
Please do not rework #33 for it.

## Gate 3 — Determinism: **PASS**

No regression. Still no `System.Random`, wall clock, `Guid.NewGuid`, `Environment.TickCount` or
`DateTimeOffset` anywhere in `src/IC2.Engine` or the tests (the only textual hit is `SaveGame`'s
doc-comment explaining why there is no timestamp). The only new ordering construct in the delta is
`JsonContract`'s tie-break, which **improves** determinism: the constructor choice is now
`OrderByDescending(parameter count).ThenBy(parameter-type signature, StringComparer.Ordinal)`, so the
one place the whole schema is derived from no longer depends on undocumented reflection order. The new
`ToHashSet` in `NoHardcodedConstantsTests` is membership-only. Round-trip byte-identity still proven
for all four kinds.

## Gate 4 — Scope: **PASS**

Still exactly the 30 files of the declared Owns list — `git diff --name-status main...HEAD` is
30 × `A`, and the rework commit modified 15 of them and added none. Nothing under `data/worlds/` or
`data/rulesets/` beyond the two toy files, so T29 is still unobstructed. Confirmed untouched:
`docs/build-orchestration-plan.md` and all of `docs/`, `IC2.sln`, `Directory.Build.props`,
`.github/workflows/`, `src/IC2.Data`, `src/IC2.Inspect`, `src/IC2.Cli`, `godot/`, and
`tests/IC2.Engine.Tests/IC2.Engine.Tests.csproj`. The two model additions (`SiegeRules`,
`NavalCombatRules`) are inside `src/IC2.Engine/Model/**` and are data-only.

## Gate 5 — Correctness: **PASS**

I verified every fix with my own probes against the built engine, in a scratch project outside the
repository, rather than relying on the new tests.

**5.1 calendar — fixed at the root.**

```
start week=1 season=0 year=270
cycle: 1,3,5,7,9,11,1,3   reaches 11: True
```

`GameStateFactory` now reads all four calendar values from the ruleset, and
`ScenarioLoadingTests.The_initial_calendar_comes_entirely_from_the_loaded_ruleset` pins it with an
`Assert.NotEqual(WeekStep, Week)` that would fail if the old improvisation came back. `ToyFixtures`
genuinely no longer masks it: the mid-game calendar is now `AdvanceWeeks(initial.Calendar, …, 19)`
derived from the start rather than a hand-written `CalendarState(7, 2, 269, 19)`, so a wrong start
would now propagate into every fixture-based test instead of being overwritten. Confirmed by reading
the diff.

**`ValidateCalendar` is correct for step values generally, not just for step 2.** This was the part
worth checking rather than assuming, so I compared it against brute-force truth across six
start/step/boundary combinations including odd steps and a non-default boundary:

```
start=1 step=3 boundary=11: reachable=False accepted=False OK
start=2 step=3 boundary=11: reachable=True  accepted=True  OK
start=0 step=4 boundary=8 : reachable=True  accepted=True  OK
start=1 step=5 boundary=11: reachable=True  accepted=True  OK
start=3 step=2 boundary=11: reachable=True  accepted=True  OK
start=2 step=2 boundary=11: reachable=False accepted=False OK
```

Six for six. The reasoning is sound too: the orbit of `w → w + step (mod m)` has length
`m / gcd(step, m) ≤ m`, so iterating `WeekModulus` times always covers it — and the positivity and
range guards on `WeekStep`/`WeekModulus`/`SeasonAdvanceFromWeek`/`StartWeek`/`StartSeasonIndex` are
checked first, so there is no degenerate input that loops or divides badly.

**5.2 fortification — refusal semantics match the reports, not a convenient reading.**

```
code 340 (40% + 3 pending): canPlace=False  maxOrderable=0
code 100 (at maximum)     : canPlace=False  maxOrderable=0
radix 1000 separated from maxPercent 100: WithOrder(0,1)=1000 inProgress=True pending=1
```

I checked this against the source rather than the claim.
`decompiled-unit-map-orders-and-record-fields.md` says `TFortifyCity` is *"Refused if the city is
under siege, if fortification is already 100 ('This city cannot be fortified any further.'), or if it
is already > 100 ('This city is already being fortified.')"*. `CanPlaceOrder` reproduces exactly the
two non-siege refusals and correctly leaves the siege one to the caller, since the besieged flag lives
on the city and not in the stored word. `MaxOrderablePoints` returning 0 while pending is the faithful
consequence, not a convenience — and the updated `ModelCoverageTests` assertion is now the correct
value, with the useful addition that after `AfterSiegeAttempt` wipes the pending order the city
becomes orderable again for 60. `IsOrderInProgress` now keys off `MaxPercent`, so the radix/maximum
conflation is gone. The inherited 0%-orders-1-point ambiguity is documented as deliberately faithful,
which is what I asked for.

**5.3 state references — closed on all four fields**, not just the one I named:

```
city.owner      : rejected (UnresolvedReferenceException)
city.allegiance : rejected (UnresolvedReferenceException)
army.nation     : rejected (UnresolvedReferenceException)
fleet.nation    : rejected (UnresolvedReferenceException)
```

The embark link is now mutual in both directions with a clear message naming the disagreeing side.

**5.4 arithmetic — no path escapes untyped.** I probed the overflow case I originally found plus five
neighbours, including the `65536 × 65536` product that wraps an `int` to exactly 0:

```
overflow run      : typed (MalformedGameDataException)
negative count    : typed (MalformedGameDataException)
bad base64        : typed (MalformedGameDataException)
huge size         : typed (MalformedGameDataException)
int-overflow size : typed (MalformedGameDataException)
negative size     : typed (MalformedGameDataException)
```

The `long` + `Array.MaxLength` range check is the right shape — better than `checked`, since it keeps
the failure inside the `InvalidOperationException` channel `ValidateWorld` already funnels into a
typed error, instead of throwing an `OverflowException` past that catch. The run guard is now a
subtraction. DoD 3's contract holds for this whole class of file now, which it did not before.

**All six lower-severity notes closed**, and I re-probed the three with observable behaviour:

```
45 appends into a 40-slot log: slots=40 cursor=39 consistent=True first=m5
nested state.schemaVersion=99: typed (found=99)
LoadFile on a directory      : typed (MalformedGameDataException)
```

The news-buffer fix is better than what I asked for: `while (current.Count >= capacity)` is
self-healing rather than merely correct for consistent input, and `IsConsistent()` is enforced at load
so an inconsistent log cannot arrive in the first place. `ProvenanceMap.GetHashCode` now mixes in the
count, and `GameStateFactory`'s `<exception>` doc now describes what the method actually does.

### On `ToyFixtures.AdvanceWeeks`

Worth a sentence since it is new test-side logic that walks the confirmed calendar cycle. I checked it
against the report (`week = (week+2) mod 12`, season on the wrap departing week 11, year on the season
wrap) and it is right, including evaluating the wrap condition *before* stepping. Keeping it in the
test project with an explicit comment that the calendar engine is T06's, so it cannot be mistaken for
an implementation, is the correct call. It is in scope (`tests/IC2.Engine.Tests/Model/**`) and does
not pre-empt T06.

---

## Summary

| Gate | Round 1 | Now |
| --- | --- | --- |
| 1 DoD reproduced | FAIL (DoD 5) | **PASS** |
| 2 Provenance | FAIL (2.1) | **PASS** |
| 3 Determinism | PASS | **PASS** |
| 4 Scope | PASS | **PASS** |
| 5 Correctness | FAIL (4 findings) | **PASS** |

Both judgement calls ruled in the implementer's favour, on the evidence rather than by deference: the
narrowed DoD 5 docstring is a corrected over-claim with the underlying defect fixed at the root (backed
by my own literal sweep of both namespaces), and the conservative empty tier list with a two-sided
provenance note is the right resolution of 2.1 — **do not flip it**.

Nothing here needs a human decision. The one genuine evidence conflict in the diff — siege's `20%` vs
`10%` — is still correctly carried as two separately-named flags with the ambiguity written down and
handed to T17, so §6.6 case 4 is not tripped. The only open item is a wording nit on
`startSeasonIndex`'s `confirmed:` tag, which belongs to T29's real ruleset and is explicitly **not** a
reason to rework this PR.

Ready for the orchestrator to merge. I have not merged and have changed nothing in the tree.

PR #34, opened 2026-09-13
T04 Fixtures corpus

### LINKED ISSUE #4: T04 Fixtures corpus
**Design milestone:** M1 (fixtures half); implements `design-audit.md` §4.4 · **Branch:** `task/T04-fixtures-corpus`
**Model / effort:** Sonnet / High · **Reviewer:** **Opus / Medium**
**Owns:** `tests/fixtures/**`, `tests/IC2.Engine.Tests/Fixtures/**`

### Scope
Transcribe **every exact number in the research repo's `docs/reports/`** ([diegoami/imperial-conquest-2-research](https://github.com/diegoami/imperial-conquest-2-research), not this repo — moved there in the repo split) into one typed JSON corpus with a small loader. Each entry carries `id`, `value`, `source` (the report filename) and `tag` (`confirmed` / `derived` / `designed`, matching the report's own tagging). This is **pure transcription — nothing derived, nothing computed** — so that a later disagreement is always between code and a fixture, never between two agents' independent readings of the same report.

Minimum contents: the twelve `design-audit.md` §4.4 fixtures; the 12-entry terrain move-cost table; the full unit-type stat table; the 5×5 type-effectiveness matrix; the `+0x26` power weights (LI 20 · HI 100 · Ar 40 · LC 60 · HC 120); the diplomatic cooldowns (−8 / −24 / −18); the loyalty floors (40 / 65 / →90); the news-log message literals; the calendar constants; and the caps (100,000 troops, 20 units, 198 armies, 40 news slots, 50 mercenary slots, 1,000 purse, 990 unity, 30,000 melee, 40%).

**Any number that cannot be found in a report is omitted and reported in the PR body — never invented.**

### Done when
- [ ] A test loads the corpus and fails if any entry has an empty `value`, `source` or `tag`
- [ ] A test asserts every `source` names a file present in a committed manifest of the research repo's real report filenames (`tests/fixtures/known-reports.json` or similar, owned by this task, generated once from `gh api repos/diegoami/imperial-conquest-2-research/contents/docs/reports` — filenames only, no content vendored). Replaces the original "exists under `docs/reports/`" line, which assumed the reports were still local; they moved to [diegoami/imperial-conquest-2-research](https://github.com/diegoami/imperial-conquest-2-research) in the repo split. This stays fully offline and CI-green while still catching an invented or typo'd citation.
- [ ] A test asserts the corpus contains every id in a committed required-ids list, so a later task cannot silently find its fixture missing
- [ ] A test asserts no two entries share an id

### Blocked by
- [ ] #1 T01 Build scaffolding and CI

### Hazard
The highest-leverage task in the plan for silent error. Reviewed by Opus specifically to re-derive a sample of entries from their cited reports.

---
Full task entry (the binding contract): [`docs/build-orchestration-plan.md` → T04 Fixtures corpus](https://github.com/diegoami/imperial_conquest_2/blob/main/docs/build-orchestration-plan.md#t04-fixtures-corpus)

diegoami:
Closes #4

## Task

- **Task**: T04 Fixtures corpus — [plan entry](https://github.com/diegoami/imperial_conquest_2/blob/main/docs/build-orchestration-plan.md#t04-fixtures-corpus)
- **Branch**: `task/T04-fixtures-corpus`

## Round-1 rework applied (this update)

The Opus reviewer's round-1 verdict (`status:rework`) found **zero wrong values** across all 343 original entries — gates 1 (DoD), 3 (determinism) and 4 (scope) passed cleanly; gate 2 (provenance, a cluster of ~12 citation/tag/note defects) and gate 5 (a HIGH compile-blocking namespace collision, found in the reviewer's follow-up comment) did not. Fixed in full:

- **C1 (HIGH)** — `Fixtures` collided with its own declaring namespace `IC2.Engine.Tests.Fixtures`, so `Fixtures.Get(...)` failed `CS0234` from any sibling namespace — exactly the call shape plan §2.4 promises every downstream task. Renamed `Fixtures` → `FixtureCorpus`. Added `DownstreamNamespaceUsageTests.cs`, declared under a sibling namespace (`IC2.Engine.Tests.Calendar`, standing in for a real downstream task's own subfolder) while staying physically inside this task's Owns list, proving `FixtureCorpus.Get(...)` now resolves unqualified from outside `IC2.Engine.Tests.Fixtures`.
- **P1** — `movement.blockingMarkerRange.city` was `"20-199"` cited to a report containing no marker ranges at all, silently reconciling two different claims. Split into `movement.blockingMarkerRange.city` = `"20-99"` (the movement-*blocking* range, `terrain-move-cost-table-in-dat.md`) and a new `mapMarker.cityCodeRange` = `"20-199"` (the map-marker *identity* range, `rivers-and-map-markers.md`) — carried as a genuine, unreconciled conflict, not picked a winner.
- **P2** — re-pointed `movement.blockingMarkerRange.{army,fleet}` and `movement.fleetSeaOnlyCodes` to `terrain-move-cost-table-in-dat.md`, which actually contains them.
- **P3** — re-pointed `battle.naval.dockedFleetCannotBeAttacked`'s source to `decompiled-diplomacy-peace-terms-and-instant-battles.md` (the report with the exact error text), off a source with no naval content.
- **P4** — the `×9/10` defender-penalty claim in the old `capture.attackerDefenderPenaltyAmbiguity` entry exists in **no research report** (grep-confirmed across all 46) — it lives only in `docs/design-audit.md` §2.13. Per the plan's rule ("a number that cannot be found in a report is not invented — it is omitted and reported in the PR body"), **omitted from the corpus entirely**; see the omissions list below. `capture.nonAllegiantDefenderPenalty` (the correctly-sourced −20% half) is untouched.
- **P5** — retagged `victory.yearLimitBC` `confirmed` → `derived`; the report's own "What this does not establish" explicitly lists whether 250 BC is the real end-of-game trigger as open.
- **P6** — fixed `fleet.marker.owner{1,3}Ships{90,70}`'s notes, which repeated a Cales/Capua home-port reading the source report's own correction header retracts; corrected to Carthage/Ptolemaic owner attribution.
- **P7** — re-pointed `calendar.nationCount` from a hotseat-player-count line (`impconq2-initial-report.md`) to the 16-entry `turnOrderTable` in `decompiled-turn-and-calendar-sequencing.md`.
- **P8** — restored four string literals verbatim: `newsMessage.fleetFinished`'s dropped `[fleet[+20]]` index; `error.tradeRefusedCooldown` / `error.tradeNotAllowedRelation` / `error.peaceRefused`'s backtick-`X` placeholder, previously rewritten to `{X}`.
- **P9** — transcribed the two previously-skipped reports: `army-to-army-transfer-confirmed.md` (exact reciprocal 27,705-troop / 70-money conservation — the direct twin of the 79-ton §4.4 fixture) and `field-recruitment-uniform-attrition-and-fleet-drift.md` (the 2.50%–2.85% army-wide attrition band, tagged `derived` per the report's own hedge).
- **P10** — this PR body's counts are now regenerated programmatically (script below), not typed by hand.
- **Additional fixes from the reviewer's follow-up comment**: `terrain.daTOffset` → `terrain.datOffset` (id typo); dropped two duplicate-fact-under-two-ids pairs (`recruitment.armyCap100k.errorString`, `battle.melee.rawLossAbsoluteCap`), each replaced by an `.evidence`-suffixed id holding only the supporting text, not a second copy of the number; `required-ids.json` widened to guard all `newsMessage.*` and `error.*` ids (was 13 of 17 / 0 of 20, now 17/17 and 20/20); `BuildIndex` now throws naming the duplicate id instead of silently last-winning (its comment previously claimed this and didn't do it).
- **The five LOW correctness nits** (gate 5): `IsEmptyValue`'s string check now uses `IsNullOrWhiteSpace`, consistent with `Id`/`Source`/`Tag`; added an id-non-empty check to the same test; `LoadCorpus`'s per-entry parse now wraps a missing/null required property with the array index and id for a diagnosable message; the `JsonElement` record-equality trap on `FixtureEntry` is documented (not silently "fixed" with unproven machinery) since nothing uses entry-to-entry equality today; `FindRepositoryRoot`'s duplication of T02's `TestPaths` is unchanged — both are one-task-owned files with no shared location to hoist to yet, matching the reviewer's own "worth a follow-up issue" framing rather than "fix now."

```text
$ python3 -c "
import json
from collections import Counter
corpus = json.load(open('tests/fixtures/corpus.json', encoding='utf-8'))
entries = corpus['entries']
print('total:', len(entries))
print(Counter(e['tag'] for e in entries))
"
total: 355
Counter({'confirmed': 347, 'derived': 8})
```

**Corrected counts: 355 entries — 347 `confirmed`, 8 `derived`, 0 `designed`** (was incorrectly reported as 343 / 339 / 4 in the pre-rework body). The 8 `derived` entries, each hedged by its own source report: `economy.rebellionLoyaltyThreshold`, `mobilization.newArmyTemplate`, `matrix.orientationCaveat`, `battle.tactical.adjacencyPromotionRule`, `economy.aiStabilityCheckChance`, `newsMessage.pendingDiplomaticOfferParaphrase`, `victory.yearLimitBC` (new this round, P5), `attrition.armyWideUniformLossBand` (new this round, P9).

| --- | --- |
| `tests/fixtures/corpus.json` | `tests/fixtures/**` |
| `tests/fixtures/known-reports.json` | `tests/fixtures/**` |
| `tests/fixtures/required-ids.json` | `tests/fixtures/**` |
| `tests/IC2.Engine.Tests/Fixtures/FixtureEntry.cs` | `tests/IC2.Engine.Tests/Fixtures/**` |
| `tests/IC2.Engine.Tests/Fixtures/FixturePaths.cs` | `tests/IC2.Engine.Tests/Fixtures/**` |
| `tests/IC2.Engine.Tests/Fixtures/FixtureCorpus.cs` (renamed from `Fixtures.cs`, C1) | `tests/IC2.Engine.Tests/Fixtures/**` |
| `tests/IC2.Engine.Tests/Fixtures/FixturesCorpusTests.cs` | `tests/IC2.Engine.Tests/Fixtures/**` |
| `tests/IC2.Engine.Tests/Fixtures/DownstreamNamespaceUsageTests.cs` (new, C1 proof) | `tests/IC2.Engine.Tests/Fixtures/**` |

- [x] No change to `docs/build-orchestration-plan.md`, original game files, `assets.local.ini`, `.gitignore`'s exclusion policy, or anything under `docs/reports/`. `tests/IC2.Engine.Tests/IC2.Engine.Tests.csproj` (T01's file) remains untouched.

## DoD evidence

```text
DoD 1: A test loads the corpus and fails if any entry has an empty `value`, `source`, or `tag`.
$ dotnet test IC2.sln --filter "FullyQualifiedName~NoEntryHasAnEmptyValueSourceOrTag"
  Passed IC2.Engine.Tests.Fixtures.FixturesCorpusTests.NoEntryHasAnEmptyValueSourceOrTag [7 ms]
Total tests: 1, Passed: 1

DoD 2: A test asserts every `source` names a file present in a committed manifest of the
research repo's real report filenames (offline).
$ dotnet test IC2.sln --filter "FullyQualifiedName~EverySourceNamesAFileInTheKnownReportsManifest"
  Passed IC2.Engine.Tests.Fixtures.FixturesCorpusTests.EverySourceNamesAFileInTheKnownReportsManifest [13 ms]
Total tests: 1, Passed: 1

DoD 3: A test asserts the corpus contains every id in a committed required-ids list.
$ dotnet test IC2.sln --filter "FullyQualifiedName~CorpusContainsEveryRequiredId"
  Passed IC2.Engine.Tests.Fixtures.FixturesCorpusTests.CorpusContainsEveryRequiredId [10 ms]
Total tests: 1, Passed: 1

DoD 4: A test asserts no two entries share an id.
$ dotnet test IC2.sln --filter "FullyQualifiedName~NoTwoEntriesShareAnId"
  Passed IC2.Engine.Tests.Fixtures.FixturesCorpusTests.NoTwoEntriesShareAnId [7 ms]
Total tests: 1, Passed: 1

C1 fix proof: FixtureCorpus.Get(...) resolves unqualified from a sibling namespace.
$ dotnet test IC2.sln --filter "FullyQualifiedName~DownstreamNamespaceUsageTests"
  Passed IC2.Engine.Tests.Calendar.DownstreamNamespaceUsageTests.FixtureCorpusIsCallableUnqualifiedFromASiblingNamespace [13 ms]
Total tests: 1, Passed: 1

Full build and suite:
$ dotnet build IC2.sln
Build succeeded.
    0 Warning(s)
    0 Error(s)

$ dotnet test IC2.sln
Passed!  - Failed: 0, Passed: 1, Skipped: 0, Total: 1  - IC2.Data.Tests.dll (net10.0)
Passed!  - Failed: 0, Passed: 66, Skipped: 0, Total: 66 - IC2.Engine.Tests.dll (net10.0)
(65 after round 0, +1 for DownstreamNamespaceUsageTests this round.)
```

All four Done-when tests were independently re-verified this round to still have teeth against the *updated* corpus (whitespace-only value, duplicate id via `Get()`, and a missing-property entry all reproduced and correctly failed with clear diagnostics, then reverted — working tree clean before commit).

## Provenance checklist

- [x] Every constant introduced or changed traces to a `tests/fixtures` entry — this PR *is* the fixtures corpus, so every entry traces to a `source` report filename instead (checked by DoD 2).
- [x] Every `[designed]` value states what was searched and came up empty. **0 of 355 entries is tagged `designed`.**
- [x] Nothing was invented to fill a gap: a value that could not be found in a report is omitted and listed below, not guessed.

**Evidence conflicts carried, not resolved** (per this task's explicit instruction):
- `movement.blockingMarkerRange.city` (20–99, the movement-blocking range) vs. `mapMarker.cityCodeRange` (20–199, the map-marker identity range) — two different reports, two different claims about a city's code range, genuinely disagreeing, neither reconciled. **New this round (was previously a single silently-reconciled entry — the round-1 reviewer's P1 finding).**
- `capture.nonAllegiantDefenderPenalty` (−20%, `decompiled-city-capture-resolution.md`, confirmed) stands alone now — its former conflict partner (the ×9/10 claim) is omitted per the list below rather than carried, since that half has no report source at all (a genuine conflict needs two *sourced* readings; this only ever had one).
- `reparation.ptolemaicDelta` (−2,269, clean) alongside `reparation.greeceToMacedonia.notedAmount` (998, whose actual treasury deltas that turn don't match cleanly) — unchanged from round 0.
- `mercenary.felsina.quarterlyCostTalents` (51) is the mercenary's recurring quarterly upkeep, not the one-time hire cost `mercenary.hireCostFormula` describes — the round-1 reviewer independently re-derived this and confirmed the reading is correct.

**Values omitted because no source was found in any research report** (never invented):
- **The ×9/10 (−10%) defender-strength reduction "when the attacking nation equals the city's allegiance"** — `docs/design-audit.md` §2.13 attributes this to `FUN_0044b27c`, but grepping all 46 reports for `9/10`, `× 9`, `-10%` and variants finds it in **no report**; `decompiled-city-capture-resolution.md`'s own `FUN_0044b27c` pseudocode (which this task did read and did transcribe from) contains no such term. Per the plan's rule, omitted rather than sourced to a document outside `docs/reports/`. **Not lost**: T02 already ships this ambiguity as a named ruleset flag with the disagreement recorded in its own `_provenance`, so T17 still inherits it from there.
- Save-file/DAT record byte offsets and strides (army 656B, fleet 26B, city 34B, nation 1,172B, unit slot 32B, etc.) — `IC2.Data`'s parsing concerns, already implemented there.
- SHA-256 hashes and file sizes of saves/screenshots/recordings — file-integrity bookkeeping, not gameplay data.
- Screenshot pixel/owner color-code table (`rivers-and-map-markers.md`) — display-only.
- The shooting-range formula's exact constants (`FUN_0043845c`) and the melee defense-factor formula's exact constants (`FUN_00438420`) — both explicitly "not yet decompiled" in every report that mentions them.
- Seasonal population-growth-rate and army-supply-consumption-rate table values — addresses named, contents "weren't extracted this pass" (`decompiled-turn-and-calendar-sequencing.md`).
- The quarterly treasury income formula's three additional terms beyond the tax-income term (a mobilization-linked bonus `/4`, a flat upkeep-style cost `×7`, a wealth-proportional drain `/20000`, plus one wholly undecompiled income helper) — fields referenced only by unidentified offset, no numeric cross-check performed.
- The population/fortification-loss smoothing formula's exact target/rate (`FUN_0044b230`) — explicitly "not pinned down this pass."
- Nation-elimination-cascade (`FUN_0044c528`) trigger thresholds and effects beyond what's already captured — "not decompiled in depth."
- The siege defender-strength formula's "third field" identity and its ×5/3 fortification-bonus gating condition — both explicitly unidentified; the formula *shape* is transcribed (`capture.siegeDefenderStrengthFormula`), the missing piece is not guessed at.
- `unitType.*.field0x20`'s semantic meaning — the five raw values ARE transcribed (18/2/18/15/4), each note stating the meaning is unidentified; no name invented for the field.

## Determinism

- [x] This PR touches no gameplay path — static JSON data plus a read-only xunit test-project loader (`System.Text.Json.JsonDocument.Parse`, `List<T>` built in file order). No `System.Random`, wall-clock reads, `Guid.NewGuid`, or hash-based iteration order anywhere in the new code. `FixtureCorpus.All` preserves `corpus.json`'s own array order; `BuildIndex()`'s duplicate-key behavior is now a hard throw (not silently relied on), and DoD 4's own test enumerates `FixtureCorpus.All` directly via `GroupBy`, never the dictionary index, so a duplicate can't be silently swallowed by either path.

## Notes for the reviewer

- **`battle.romeGaul.*` and `battle.observation.*`** remain two *different* Rome-vs-Gaul battles from two different report pairs, kept under clearly distinct id prefixes.
- **T16's hazards** call out that the type-effectiveness matrix, the 40% melee cap, and the tactical morale array must not appear in *T16's* diff. They deliberately DO appear here, tagged and noted as reserve/tactical-path data — T04's job is to hold the full evidence base, not pre-filter what a later task may consume.
- I disagree with one framing in the reviewer's LOW list, with reasoning rather than silent skipping: `FindRepositoryRoot`'s duplication of T02's `TestPaths` is left as-is. The reviewer's own note called it "worth a follow-up issue... once some task owns a test-support folder" rather than something to fix now, and no task currently owns such a folder — hoisting it here would be an out-of-scope structural change this task's Owns list doesn't cover.
- The `error.fleetCannotAttackDocked` / `battle.naval.dockedFleetCannotBeAttacked` pair (flagged as "the same shape as P3") is left as two separate ids after P3's source fix: they hold genuinely different value shapes (the exact UI string vs. a boolean mechanic flag) documenting the same fact from two angles, not a literal duplicate the way the two dropped pairs were (identical value AND source) — so no consistency test was added for it, only the source correction.


diegoami:
## T04 review — Opus reviewer, five gates

**Verdict: request changes (`status:rework`).** Gates 1, 3, 4 and 5 pass cleanly. Gate 2 (provenance) does not.

The headline is good news and should scope the fix: **I found zero wrong values.** I ran an automated verbatim check of all 214 numeric values and all 115 string values against their cited report's raw text, then hand-re-derived ~95 entries line-by-line from 20 of the 46 reports. Every number in this corpus exists, in the form given, somewhere in the research repo. The twelve `design-audit.md` §4.4 fixtures, the 12-entry terrain table, the full unit-type stat table, the 5×5 matrix, the `+0x26` power weights, the diplomatic cooldowns, the loyalty floors, the calendar constants, the news-log literals and all nine caps are present **and correct**.

What fails gate 2 is a cluster of **~12 citation, tagging and note defects** — `source` fields naming a report that does not contain the claim (the exact `design-audit.md` §2.7 failure mode this task exists to prevent), one silently reconciled value conflict, two notes repeating a claim the source report itself retracted, and one `confirmed` tag the source report explicitly lists under "What this does not establish". All are patchable in place; **the corpus does not need redoing.**

---

## Gate 1 — DoD, reproduced by me. **PASS**

Reproduced on `be4462c`, clean tree, Debug and CI-equivalent Release:

```
$ dotnet build IC2.sln                 → Build succeeded. 0 Warning(s) 0 Error(s)
$ dotnet test  IC2.sln                 → 65 passed (Engine) + 1 passed (Data)
$ dotnet build IC2.sln -c Release      → Build succeeded. 0 Warning(s) 0 Error(s)
$ dotnet test  IC2.sln --no-build -c Release → 65 + 1 passed
```

65 engine + 1 data confirmed (T02 left it at 59+1; +6 here). Zero warnings under `TreatWarningsAsErrors`.

**Teeth — verified by me, not taken from the PR body.** I injected defects into `corpus.json`, ran the suite, and restored:

| Injected defect | Result |
| --- | --- |
| `source` → `mercenary-pool-records.md` (near-miss typo of a real filename) | `EverySourceNamesAFileInTheKnownReportsManifest` **failed**: `Sources not present in known-reports.json: mercenary-pool-records.md` |
| deleted the `caps.maxArmies` entry | `CorpusContainsEveryRequiredId` **failed**: `Required ids missing from the corpus: caps.maxArmies` |
| duplicated id `caps.maxNewsSlots` | `NoTwoEntriesShareAnId` **failed**: `Duplicate fixture ids: caps.maxNewsSlots` |
| `"value": ""` + `"tag": "  "` on `caps.maxUnity` | `NoEntryHasAnEmptyValueSourceOrTag` **failed**, and `EveryTagIsOneOfTheThreeConventionValues` **failed** too |

All four Done-when tests have real teeth, including the deliberately adversarial near-miss filename. Working tree restored and clean.

**DoD 2 is genuinely offline.** No `HttpClient` / `WebClient` / `System.Net` / `Process` anywhere in `tests/IC2.Engine.Tests/Fixtures/**`; the only occurrence of `gh api` is inside an XML doc comment. The test reads `tests/fixtures/known-reports.json` off disk and nothing else. I separately verified the manifest against the live source — `gh api repos/diegoami/imperial-conquest-2-research/contents/docs/reports` returns exactly 46 filenames, and the manifest's 46 entries are byte-identical to that set with zero adds or drops (also identical to the local read-only copy). Manifest is sorted, so a future regeneration diffs cleanly. ✔

`required-ids.json`: 146 ids, no duplicates, all present in the corpus, and covering every item in the plan's "minimum contents" line.

---

## Gate 2 — Provenance. **FAIL**

### Method

1. Automated verbatim pass: every one of the 214 numeric values checked for its literal presence (decimal, comma-grouped, and hex forms) in its cited report — **0 misses**. Every one of the 115 string values checked verbatim, then whitespace-normalised — 68 flagged, of which 64 are legitimate one-line summaries of a report's multi-line pseudocode block (`reparation.formula`, `battle.naval.winnerDamageFormula`, etc., all verified by hand against the block they summarise) and 4 are genuine findings, below.
2. Hand re-derivation of ~95 entries from 20 reports read in full: `unit-type-stat-table-in-dat`, `combat-type-effectiveness-matrix`, `terrain-move-cost-table-in-dat`, `decompiled-army-movement-and-river-cost`, `rivers-and-map-markers`, `decompiled-unit-map-orders-and-record-fields`, `decompiled-diplomacy-peace-terms-and-instant-battles`, `decompiled-city-capture-resolution`, `decompiled-defection-and-siege-attrition`, `decompiled-quarterly-billing-and-economy`, `decompiled-turn-and-calendar-sequencing`, `decompiled-fleet-tax-and-mercenary-formulas`, `decompiled-combat-formula-structure`, `battle-recording-melee-cap-confirmed`, `decompiled-weather-events`, `decompiled-news-log-identified`, `mercenary-pool-record`, `galatia-elimination-and-city-resupply-confirmed`, `full-battle-resolution-rome-vs-gaul`, `battle-quality-promotion-and-morale-array-decompiled`, plus targeted greps across the other 26.

### What checked out (worth stating explicitly, since it is most of the corpus)

- **Unit-type stat table**, all 40 values, exact match to `unit-type-stat-table-in-dat.md`'s table (moves 4/2/4/6/5 · battalion 15000/6000/3500/7000/2500 · shots 7/0/25/9/0 · range 1/0/2/1/0 · `+0x20` 18/2/18/15/4 · initial 2/20/4/15/30 · quarterly 1/2/1/3/4).
- **`+0x26` power weights** LI 20 / HI 100 / Ar 40 / LC 60 / HC 120 — I specifically hunted this for the over-tagging pattern the T02 reviewer found on the `+44` season byte. It holds: `unit-type-stat-table-in-dat.md` leaves `+0x26` unidentified, but `decompiled-unit-map-orders-and-record-fields.md:36` identifies it flatly and unhedged ("`powerWeight` is unit-type-table field **`+0x26`** (LI 20, HI 100, Ar 40, LC 60, HC 120) … It is a per-type combat-power weight"). `confirmed` is correct, and the entries cite the right report.
- **5×5 matrix**, all 25 values exact; and the orientation caveat is correctly split out as a separate `derived` entry (`matrix.orientationCaveat`) rather than being folded into the values — the report says "only the axis labeling is uncertain, not the 25 values themselves", so `confirmed` on the 25 is right.
- **Terrain table**, all 12 costs + `0x1F622` exact; river shape labels in the notes (6=EW, 7=NS, 8=NE, 9=ES, 10=SW, 11=NW) match `rivers-and-map-markers.md:11-18` exactly.
- **News-log and error literals**: 13 of 17 `newsMessage.*` and 16 of 20 `error.*` match their report **byte-for-byte**, including the awkward spacing that matters (`"{CityName}   ({OldOwner})  falls to {NewOwner}."` three-then-two spaces; `"    <loser>  ends all current alliances."` leading four spaces and internal double space) and both original-game typos preserved (`conquerred`, `suuplies`). That is exactly right for a corpus later tasks assert against.
- **Cooldowns** −8/−24/−18, **loyalty floors** 40/65/→90, **calendar** +2 mod 12 / wrap at 1 / year at season 0 / 16-entry table / StateCode +2 cap 24, **all nine caps** (100,000 · 20 · 198 · 40 · 50 · 1,000 · 990 · 30,000 · 40%) — all verified against their reports.
- **All twelve `design-audit.md` §4.4 fixtures** present and correct: 442 · 482/48,173→100% · 20/34/65 · 2,440 at 15/20% · 100/5,000/30 · 333/335 · the 12-entry terrain table · Felsina 6,438 "very good" + 0xFFFF · 79-ton reciprocal · 85,000→70,000 and Masada 49,800 · 999→−1270 (Δ−2269) · 9 cities / 2 falls-to / 7 defects.
- **The three carried evidence conflicts are genuinely carried**, with distinct ids, distinct sources and explicit notes, no winner picked. `capture.nonAllegiantDefenderPenalty` sits alongside `capture.attackerDefenderPenaltyAmbiguity`; `reparation.ptolemaicDelta` alongside `reparation.greeceToMacedonia.notedAmount` (whose note correctly records Greece −1,576 / Macedonia +2,292 from `diplomatic-reparations-and-more-captures.md:25`); `mercenary.felsina.quarterlyCostTalents` is correctly identified as **upkeep, not hire cost** — I re-derived this independently and it is right. `decompiled-fleet-tax-and-mercenary-formulas.md:57` calls it "the 51-quarterly Felsina hire" and states the hire-cost formula "wasn't solved for concrete table values against" it, and `(6438/200)×1×8/5 = 32×8/5 = 51` under integer division. Good catch by the implementer; the PR body's request for a second opinion is answered — **your reading is correct.**
- **The `derived` tags are honest.** `economy.rebellionLoyaltyThreshold`, `economy.aiStabilityCheckChance`, `matrix.orientationCaveat`, `battle.tactical.adjacencyPromotionRule` (correctly `derived` per `design-audit.md` §2.10), `newsMessage.pendingDiplomaticOfferParaphrase`, `mobilization.newArmyTemplate` — each matches a hedge its own report states.

### Findings (must fix)

**P1 — `movement.blockingMarkerRange.city` = `"20-199"`: wrong source *and* a silently reconciled conflict.** The cited report, `decompiled-army-movement-and-river-cost.md`, contains **no marker ranges at all** (I read it in full; its correction header mentions "cell codes ≥ 12 (city/army/fleet markers) block the walk entirely" but names no numbers). The only report that states the *blocking* range is `terrain-move-cost-table-in-dat.md:68`:

> "**Codes ≥ 12 are not steppable at all** by this walk — a city (**20–99**), army (200–247) or fleet (300–347) marker in the path aborts the move"

`20-199` comes from `rivers-and-map-markers.md:24` — "exactly **334 cells** have values from `20` through `199`" — which is the *map-marker* range for cities, a different claim from the movement-blocking range. Two reports give two different city ranges and the corpus has quietly shipped one of them under a citation that supports neither. This is plan §6.6 case 4: carry both as distinct ids with distinct sources and a note, don't reconcile.

**P2 — `movement.blockingMarkerRange.army` (`200-247`), `movement.blockingMarkerRange.fleet` (`300-347`), `movement.fleetSeaOnlyCodes` (`[0, 1]`): correct values, wrong `source`.** None of the three appears in `decompiled-army-movement-and-river-cost.md`. `200-247`/`300-347` are in `terrain-move-cost-table-in-dat.md:68` and again in `decompiled-unit-map-orders-and-record-fields.md:157` ("`TUnitMap_SelectUnit` accepts exactly `200..247` for armies and `300..347` for fleets"); the fleets-traverse-only-0-and-1 rule is `terrain-move-cost-table-in-dat.md:78`. Re-point the `source` fields.

**P3 — `battle.naval.dockedFleetCannotBeAttacked`: `source` is `decompiled-city-capture-resolution.md`, which contains no naval content whatsoever.** The string `"You cannot attack a fleet docked at its own city !"` is in `decompiled-diplomacy-peace-terms-and-instant-battles.md:117`. The entry's own note already admits this ("decompiled-diplomacy-peace-terms-and-instant-battles.md is the source of the exact error text") — so just make `source` say so. As written, an automated citation check passes while the citation is false, which is precisely §2.7.

**P4 — `capture.attackerDefenderPenaltyAmbiguity`: half the value's content appears in no research report at all.** I grep'd all 46 reports for `9/10`, `× 9`, `-10%` and every variant: the "×9/10 (−10%) defender reduction when the attacking nation equals the city's allegiance" claim exists **only in `docs/design-audit.md` §2.13**, not in `decompiled-city-capture-resolution.md` (whose `FUN_0044b27c` pseudocode at lines 7-15 contains no such term) nor anywhere else in the research repo. Carrying the conflict is correct behaviour and I want it kept — but the entry currently presents a design-audit claim as if the cited report contained it. Fix by making the note state explicitly that the ×9/10 half comes from `design-audit.md` §2.13 and is **not** found in any report; that is itself a useful fact for whoever re-checks it.

**P5 — `victory.yearLimitBC` = 250 tagged `confirmed` against the report's own "What this does not establish".** `decompiled-diplomacy-peace-terms-and-instant-battles.md:153` lists as unestablished:

> "Whether the 250 BC comparison is the actual end-of-game trigger or only the game-over screen's wording; the turn loop was not re-read for a year check this pass."

What is confirmed is that `THumanFalls_InitializeForm` compares `DAT_004A0332` against `250` (`0xFA`). That it is *the year limit* — which is what the id `victory.yearLimitBC` asserts and what T06/T19 will consume — is explicitly open. This is the one instance of "the tag is what the implementer judged, not what the report says" that I found. Either retag `derived`, or keep `confirmed` and rename to something that only claims the comparison (`victory.gameOverScreenYearComparand`); either way the note must carry the report's caveat, which it currently does not. (`victory.allCitiesThreshold` = 334 is fine — that one the report states affirmatively at line 144.)

**P6 — `fleet.marker.owner1Ships90` / `fleet.marker.owner3Ships70` notes repeat a claim their own report retracts.** The notes say "Fleet 0 based at **Cales** (90 ships)" and "Fleet 1 based at **Capua** (70 ships)". `fleet-order-at-caere.md`'s correction header (line 3) explicitly withdraws that: the `+20` values `97` and `100` for these two deployed fleets "are condition percentages, **not Cales and Capua**". The correct attribution is owner 1 = Carthage and owner 3 = Ptolemaic (`decompiled-unit-map-orders-and-record-fields.md:159`). The values 333/335 and the 90/70 ship counts are right; only the home-port attribution in the notes is a retracted reading. Carrying a claim a report itself retracted is the `design-audit.md` §2 pattern, and the notes are what a later task reads.

**P7 — `calendar.nationCount` = 16 cited to a player count.** `impconq2-initial-report.md:16` says "The readme calls this a Windows 95 shareware release for **1–16 players** on one computer." That is a hotseat player count, not a nation count, and it is the weakest available citation for a number that is directly confirmed three other ways: the 16-entry `turnOrderTable` (`decompiled-turn-and-calendar-sequencing.md:10`), the 16-entry relation matrix (`decompiled-diplomacy…:12`), and the "16-record nation table" (`mercenary-pool-record.md`). Re-point at one of those.

**P8 — four string literals modified in transcription.**
- `newsMessage.fleetFinished`: report has `"<nation> finishes a new fleet at <cityName[fleet[+20]]>"` (`decompiled-unit-map-orders-and-record-fields.md:43`); corpus drops the `[fleet[+20]]` index. Inconsistent with `newsMessage.conqueredByNation`, which correctly preserved `<nation[+0x44E]>`.
- `error.tradeRefusedCooldown`, `error.tradeNotAllowedRelation`, `error.peaceRefused`: the reports render the placeholder as `` `X` ``; the corpus rewrites it to `{X}`. The corpus also mixes `{CityName}` (matching `decompiled-city-capture-resolution.md`) and `<winner>` (matching the diplomacy report) elsewhere.

The mixed `<…>`/`{…}` convention is faithful to the reports and fine, but the four rewrites above are not transcription. Either restore them verbatim, or add one line to `corpus.json`'s `_comment` stating the normalisation rule, so T12/T13 know what they are asserting against.

**P9 — omission accounting in the PR body is inaccurate, and two substantive reports were silently skipped.** The PR body describes the 14 zero-contributing reports as "save-diff reports whose numbers were superseded by a later, more authoritative report … plus pure file-inventory/hash reports". That is not true of at least two:

- `army-to-army-transfer-confirmed.md:16-21` — an **exact reciprocal 27,705-troop conservation** (72,177→99,882 / 55,932→28,227, with the six transferred units summing to 27,705 exactly). This is the direct twin of `supplyTransfer.deltaTons` (79), which *is* in the corpus and *is* one of the §4.4 fixtures. Nothing supersedes it. Plus the `10s`/`100s` stepper constants.
- `field-recruitment-uniform-attrition-and-fleet-drift.md:24-43` — a **uniform army-wide attrition band of 2.50%–2.85% across twelve units**, which the report calls "the first quantitative evidence toward a battle/attrition formula". Not superseded by anything; the later combat reports cover per-exchange melee, not army-wide attrition.

The task's scope line is "transcribe **every exact number** in `docs/reports/`", and the hazard rule is that anything not found is omitted **and reported in the PR body**. These were found and neither transcribed nor reported. Add them (they're a handful of entries), or state the real reason they were left out.

**P10 — the PR body's own counts are wrong.** It says "343 total — **339** `confirmed`, **4** `derived`". The corpus is **337 confirmed / 6 derived** (`economy.rebellionLoyaltyThreshold`, `mobilization.newArmyTemplate`, `matrix.orientationCaveat`, `battle.tactical.adjacencyPromotionRule`, `economy.aiStabilityCheckChance`, `newsMessage.pendingDiplomaticOfferParaphrase`). The body's own bullet list under the "**The 4 `derived`-tagged entries**" heading has five bullets and omits `mobilization.newArmyTemplate` entirely. The tags themselves are fine — only the accounting is wrong, but this is the number the orchestrator and the next reviewer will quote.

### Suggestion (not blocking)

`required-ids.json` guards 13 of the 17 `newsMessage.*` ids but **none of the 20 `error.*` ids**. T12 (diplomacy) and T13 (fleets) will assert against those literals, and DoD 3 is the only thing stopping one from silently vanishing. Consider adding them.

---

## Gate 3 — Determinism. **PASS**

No `System.Random`, `DateTime`/`DateTimeOffset`, `Guid.NewGuid`, `Stopwatch`, `Environment.*` or `Process.*` anywhere in the new code (grep'd). `Fixtures.All` is built by `EnumerateArray()` into a `List<T>`, so enumeration order is `corpus.json`'s own file order — stable across runs and machines. `Dictionary`/`HashSet` are used only for `Contains`/`TryGetValue`, never enumerated for an order-dependent result, so the loader resolves every id identically every run and later tasks' fixtures will too. Assertion failure messages are built from the list, not the hash sets, so even the diagnostics are deterministic. `Lazy<T>` uses the default thread-safe mode; `LazyById` → `BuildIndex` → `All` → `LazyAll` is a chain, not a cycle. `ToLowerInvariant()` (not `ToLower()`) and `StringComparer.Ordinal` throughout — no culture sensitivity. ✔

## Gate 4 — Scope. **PASS**

All 7 changed files are inside the Owns list (`tests/fixtures/**`, `tests/IC2.Engine.Tests/Fixtures/**`); `git diff main...HEAD --name-only` shows nothing else. **Zero diff to `docs/`** — the stray docs commit is gone from the branch as expected, and `docs/build-orchestration-plan.md` is untouched. `tests/IC2.Engine.Tests/IC2.Engine.Tests.csproj` is untouched, confirmed by an explicit path-scoped diff. `FixturePaths.cs` reuses T02's `Model/TestPaths.cs` "walk up to `IC2.sln`" pattern faithfully, with a docstring that says why. CI-safe: `ci.yml` checks out the repo root and runs `dotnet build -c Release` / `dotnet test --no-build -c Release`, so `AppContext.BaseDirectory` is `<repo>/tests/IC2.Engine.Tests/bin/Release/net10.0/` and the walk finds `IC2.sln` four levels up — I ran the exact Release `--no-build` sequence locally and all 66 tests pass. ✔

## Gate 5 — Correctness (`/code-review --effort high`). **PASS**

Ran `/code-review --effort high` over the diff and did my own read of all ~350 new lines. No correctness bugs. `JsonDocument` is disposed inside `LoadCorpus` but `value` is `.Clone()`d first, which is exactly the right call and easy to get wrong. `BuildIndex`'s explicit loop (instead of `ToDictionary`) with its comment about not swallowing a duplicate is a good call, and DoD 4 correctly enumerates `Fixtures.All` rather than the index, so a duplicate can't hide. Nits, none blocking:

- `IsEmptyValue` uses `string.IsNullOrEmpty` for JSON strings while `Source`/`Tag` use `IsNullOrWhiteSpace`, so a `"value": "   "` would pass DoD 1. Make it consistent.
- No check that `id` itself is non-empty. DoD 1's wording doesn't demand it, but it's one predicate.
- `LoadCorpus`'s `GetProperty("id")` throws a `KeyNotFoundException` that names neither the entry nor its index if a `value`/`source`/`tag` key is *missing* (as opposed to empty) — the `FixtureEntry` docstring's promise of "a clean assertion failure instead of an exception thrown deep inside JSON deserialization" holds for empty, not for absent.
- `FixtureEntry` is a `record` containing a `JsonElement`, so synthesised value equality compares `JsonElement`'s internal cursor, not the JSON value — two entries with identical values won't compare equal. Unused today; a trap if a later task writes `Assert.Equal(a, b)`.
- `FindRepositoryRoot` is duplicated verbatim from T02's `TestPaths`. Justified — neither task owns a shared location — but worth a follow-up issue to hoist it once some task owns a test-support folder.

---

## Bottom line

343 entries, 214 numeric and 115 string values, **zero wrong values**. The corpus's substance is solid and the hard parts — byte-exact news literals, the upkeep-vs-hire-cost trap, the `+0x26` weights, honest `derived` tagging, all three conflicts carried rather than resolved — were all done right. The blockers are P1–P8 (citation/tag/note accuracy, ~10 entries) plus P9–P10 (PR-body accuracy). **Do not redo the corpus** — patch those entries and correct the PR body's counts and omission list.

diegoami:
## Correction to my review: **gate 5 also fails** — a compile-blocking defect for every downstream task

My `/code-review --effort high` run reported after I posted, with a finding I missed. I reproduced it myself before amending. **Gate 5 goes from PASS to FAIL.** The label stays `status:rework`; the verdict is unchanged but the blocker list grows by one HIGH.

---

### C1 (HIGH) — `Fixtures.cs:15`: the class name collides with its own namespace, so `Fixtures.Get(...)` does not compile from any sibling namespace

`public static class Fixtures` is declared in `namespace IC2.Engine.Tests.Fixtures`. The type name equals the last segment of its own namespace. C# simple-name lookup walks *enclosing namespace declarations* outward before consulting using-directives, so from a sibling namespace the name `Fixtures` binds to the **namespace member** `IC2.Engine.Tests.Fixtures`, not the class — and a `using IC2.Engine.Tests.Fixtures;` cannot rescue it, because using-directives are consulted only after the enclosing declarations.

Reproduced on `be4462c`. Probe file at `tests/IC2.Engine.Tests/Model/ZzProbe.cs`:

```csharp
using IC2.Engine.Tests.Fixtures;
using Xunit;

namespace IC2.Engine.Tests.Model;

public class ZzProbe
{
    [Fact]
    public void CanCallFixturesGetFromASiblingNamespace()
    {
        Assert.Equal(2440, Fixtures.Get("tax.nationTaxBaseRome").AsInt());
    }
}
```

```
$ dotnet build IC2.sln
tests\IC2.Engine.Tests\Model\ZzProbe.cs(11,28): error CS0234: The type or namespace name 'Get'
  does not exist in the namespace 'IC2.Engine.Tests.Fixtures' (are you missing an assembly reference?)
Build FAILED.
```

That is the most natural possible call site: a later task's tests living in `IC2.Engine.Tests.Calendar` / `.Economy` / `.Combat` — the layout `IC2.Engine.Tests.csproj`'s own comment prescribes ("each later task under `src/IC2.Engine` owns its own subfolder of `tests/IC2.Engine.Tests`") — writing exactly the call the plan §2.4 promises: *"Later tasks assert against `Fixtures.Get`"*. The shipped tests pass only because `FixturesCorpusTests` is itself inside the colliding namespace, so `Fixtures` resolves to the class by the innermost-declaration rule. The defect is invisible from within this PR and lands on T06 and every task after it.

**I checked what still works, so the fix can be chosen deliberately** — a second probe compiled clean:

```csharp
using FixtureCorpus = IC2.Engine.Tests.Fixtures.Fixtures;   // alias: works
...
IC2.Engine.Tests.Fixtures.Fixtures.Get("tax.nationTaxBaseRome")  // fully qualified: works
FixtureCorpus.Get("caps.maxUnity");                              // works
FixtureEntry / FixtureTag / FixturePaths                         // no collision, work bare
```

So it is not unfixable at the call site — but it forces every one of ~14 downstream tasks to discover the CS0234 and independently invent a workaround, which is exactly the per-task rediscovery cost this corpus exists to remove. Only `Fixtures` collides; the three sibling types are fine.

**Fix:** rename the class to `FixtureCorpus` (my preference — it reads better at the call site and leaves the folder/namespace layout matching every other task's) or rename the namespace. Either is a one-file change plus the six call sites in `FixturesCorpusTests`. Probe deleted; tree clean; 65/65 still pass.

---

### C2 — independent confirmation of P1/P2

The review agent reached my finding P1 separately and from the reports themselves: `movement.blockingMarkerRange.city` = `20-199` cites a report containing no marker ranges, silently reconciles `20–99` (`terrain-move-cost-table-in-dat.md:68`) against `20-199` (`rivers-and-map-markers.md:24`), and Done-when 2 cannot catch it because the filename is in the manifest. It adds the consequence I should have spelled out: **T09 consumes this, and whether codes 100–199 block a path is a real behavioural difference**, not a cosmetic citation nit. Treat P1 as the highest-priority provenance fix.

---

### Additional findings, all verified

- **`corpus.json:74` — id `terrain.daTOffset` has a stray capital T** (should be `terrain.datOffset`). Ids are runtime string keys, and this one is not in `required-ids.json`, so a later task typing the natural spelling gets a `KeyNotFoundException` at test time rather than a compile error.
- **`required-ids.json:125` — the news block guards 13 of the corpus's 17 `newsMessage.*` ids.** The four unguarded ones are `newsMessage.endsTradingAgreements`, `newsMessage.endsAlliances`, `newsMessage.allyPeaceAgreement`, `newsMessage.pendingDiplomaticOfferParaphrase` (confirmed by set-difference). The plan's minimum contents says "the news-log message literals" and T10's DoD asserts over the corpus's news section, so these four are precisely the ones an edit can delete without Done-when 3 noticing. This upgrades my earlier non-blocking suggestion: add these four, and consider the 20 `error.*` ids too.
- **Three facts are carried under two ids each with nothing cross-checking them** — `recruitment.armyCap100k.errorString` / `error.armyTooLarge100k` (identical string *and* source), `battle.melee.rawLossAbsoluteCap` / `caps.meleeLossAbsoluteCap` (30000, same source), `mercenary.poolSlotCount` / `caps.maxMercenarySlots` (50, same source). I confirmed these three are genuinely the same fact twice, not coincidental value collisions (the corpus has plenty of the latter — six different matrix cells hold `4` — which are fine). In a corpus whose entire purpose is single-source-of-truth, a corrective edit to one leaves the other stale. Either drop one of each pair, or add a test asserting the pairs stay equal. Note this is the same shape as my P3: `battle.naval.dockedFleetCannotBeAttacked` and `error.fleetCannotAttackDocked` carry one fact under two ids **with two different `source` values**, one of which is wrong.
- **`Fixtures.cs:73` — `BuildIndex`'s comment is contradicted by its code.** The comment says the explicit loop (rather than `ToDictionary`) makes a duplicate id "fail clearly"; `index[entry.Id] = entry` silently last-wins. In a run where `NoTwoEntriesShareAnId` is already failing, every other test quietly reads the later duplicate. Fix the comment or make the loop throw with both ids named.
- **`Fixtures.cs:63` — the `KeyNotFoundException` from a *missing* key is cached by `Lazy<T>`.** I flagged that `GetProperty("id")` throws without naming the entry; the sharper version is that `Lazy<T>`'s default `ExecutionAndPublication` mode caches the exception and rethrows it on every subsequent access, so one malformed entry produces four opaque identical failures rather than one diagnosable one — the exact hazard `FixtureEntry`'s XML doc claims to have designed around.

The review run also independently re-verified, from the reports, all 25 matrix values, all 40 unit-type stat values including the `+0x26` weights, all 12 terrain costs, the 2,440 tax base against 366@15%/488@20%, the 442-talent sum, Masada 49,800, the Rome 85k→7k conservation, `floor(0.4·troops)+1` against all five recorded exchanges, the Rome/Gaul per-type totals, and the 46/46 manifest match — all clean, and all matching my own pass. It also reports the corpus as **337 confirmed / 6 derived**, confirming finding P10 (the PR body says 339/4).

---

### Revised gate summary

| Gate | Result |
| --- | --- |
| 1 DoD | **PASS** |
| 2 Provenance | **FAIL** — P1–P10 |
| 3 Determinism | **PASS** |
| 4 Scope | **PASS** |
| 5 Correctness | **FAIL** — C1 (HIGH), plus the LOWs above |

Fix order I'd suggest: **C1** (one rename, unblocks every downstream task), then **P1** (the only finding with a behavioural consequence), then the remaining citation/tag/note fixes P2–P8, then the PR-body corrections P9–P10 and the LOWs. Still nothing here requires redoing the corpus — 343 entries, 214 numeric and 115 string values, and between two independent passes we found **no wrong value in any of them**.

diegoami:
## T04 re-review — round-1 rework, Opus reviewer

**Verdict: approved (`status:approved`). All five gates pass.**

Every one of the eleven findings (C1 + P1–P10) is genuinely fixed, none of the fixes introduced a new error, and the 28 new-or-changed entries re-derive clean. This is not a round-2 disagreement — I have nothing left to escalate.

---

## Gate 1 — DoD, reproduced. **PASS**

On `9806284`, clean tree, Debug and CI-equivalent Release:

```
$ dotnet build IC2.sln                          → Build succeeded. 0 Warning(s) 0 Error(s)
$ dotnet test  IC2.sln                          → 66 passed (Engine) + 1 passed (Data)
$ dotnet build IC2.sln -c Release               → Build succeeded. 0 Warning(s) 0 Error(s)
$ dotnet test  IC2.sln --no-build -c Release    → 66 + 1 passed
```

66 + 1 confirmed (was 65 + 1; +1 for the C1 proof test).

**Teeth re-verified against the *changed* corpus** — the point being that the corpus moved under these tests, so round-0's teeth evidence no longer covers them. I injected four defects, ran, and restored:

| Injected | Result |
| --- | --- |
| duplicate id (`armyTransfer.deltaTroops` → `armyTransfer.deltaMoney`) | `NoTwoEntriesShareAnId` **failed** naming the id — *and* `BuildIndex` threw `Duplicate fixture id 'armyTransfer.deltaMoney' in corpus.json`, so both complementary paths its new comment claims actually fire |
| `source` → `rivers-and-map-marker.md` (near-miss typo of a real filename) | `EverySourceNamesAFileInTheKnownReportsManifest` **failed** naming it |
| `"value": "   "` (whitespace-only) on `attrition.armyWideUniformLossBand` | `NoEntryHasAnEmptyValueSourceOrTag` **failed** — this is the exact LOW that was fixed, so it doubles as a regression test for the `IsNullOrWhiteSpace` change |
| deleted `error.peaceRefused` | `CorpusContainsEveryRequiredId` **failed** naming it — which it **would not have done before this round**, since `error.*` ids were previously unguarded. The widened `required-ids.json` has real teeth, not just a longer list |

DoD 2 still genuinely offline: no `HttpClient`/`WebClient`/`System.Net`/`Process`/`Environment.*` anywhere under `tests/IC2.Engine.Tests/Fixtures/**`. Manifest re-verified: 46 filenames, still an exact set-match to the research repo's real `docs/reports/`.

`required-ids.json`: 170 ids, no duplicates, all present, **17/17 `newsMessage.*` and 20/20 `error.*`** (was 13/17 and 0/20), and no stale ids pointing at anything deleted this round.

---

## Gate 2 — Provenance. **PASS**

**28 new-or-changed entries re-derived from their reports. Zero wrong.** I also re-ran the full automated pass over all 355 entries: **0 of 355 numeric values missing from their cited report**, 0 sources outside the manifest. Non-verbatim string count went 68 → 67, and the delta is exactly what it should be — the four P8 literals and `capture.attackerDefenderPenaltyAmbiguity` dropped out, four new range/prose-shaped values came in. No regression anywhere.

Per finding:

- **P1 — verified against both source files myself, and the split is the honest reading.** `terrain-move-cost-table-in-dat.md:68` reads *"a city (20–99), army (200–247) or fleet (300–347) marker in the path aborts the move"* — that is the movement-**blocking** range, and `movement.blockingMarkerRange.city` = `"20-99"` now cites it. `rivers-and-map-markers.md:24` reads *"exactly **334 cells** have values from `20` through `199`, and all 334 coincide with the parsed city coordinates. These values are city map markers"* — that is the marker-**identity** range, and the new `mapMarker.cityCodeRange` = `"20-199"` cites it. These are genuinely two different claims about two different things, not one claim read two ways, and both notes say so explicitly and cross-reference each other. This is the right outcome — better than carrying them as a conflict, because there is no conflict once the two claims are stated precisely. T09 now gets `20-99` for path-blocking, which is the correct semantics for what it consumes.
- **P2 / P3 / P7 — each new source really contains the claim.** `200-247`/`300-347` and the fleets-traverse-only-0-and-1 rule are in `terrain-move-cost-table-in-dat.md` (lines 68 and 78) ✔. `"You cannot attack a fleet docked at its own city !"` is in `decompiled-diplomacy-peace-terms-and-instant-battles.md:117` ✔. The 16-entry `turnOrderTable` is in `decompiled-turn-and-calendar-sequencing.md:10` ✔.
- **P4 — deletion is complete and correctly reported.** `capture.attackerDefenderPenaltyAmbiguity` is gone; the only remaining strings matching the deleted ids anywhere in `tests/fixtures/**` are two `note` fields explaining the rework, not dangling id references. `capture.nonAllegiantDefenderPenalty` is byte-identical to before ✔. The omissions list states the deletion, the grep evidence, and the reason. **The schema-gap ruling was followed** — no `secondarySource` field was added. I also verified the body's load-bearing claim that nothing is lost: `data/rulesets/toy-ruleset.json:334-347` (T02, already on main) carries **both** readings as separate fields with a `_provenance` note that cites `docs/design-audit.md` §2.13 by name for the ×9/10 branch. That is the correct home for a design-audit-only claim — a ruleset provenance string, where citing outside `docs/reports/` is legitimate — rather than a corpus `source` field that DoD 2 constrains to report filenames. Good resolution.
- **P5** — `victory.yearLimitBC` is `derived`, and the note now quotes the report's own caveat verbatim ✔.
- **P6** — both fleet-marker notes corrected to Carthage (owner 1) / Ptolemaic (owner 3), each explicitly flagging that the source report's correction header retracts the Cales/Capua reading ✔.
- **P8 — all four literals byte-exact, checked by substring against the raw report text, not by eye:** `newsMessage.fleetFinished` (`[fleet[+20]]` restored), `error.tradeRefusedCooldown`, `error.tradeNotAllowedRelation`, `error.peaceRefused` (backtick-`X` restored) — all four `True` ✔. Nothing else in the literal set was touched, so the awkward spacing and the two original-game typos (`conquerred`, `suuplies`) survive intact.
- **P9 — all 12 new entries re-derived, 0 wrong.** Against `army-to-army-transfer-confirmed.md`: army 0 troops 72,177→99,882 and army 2 55,932→28,227 (line 16-17) ✔; Δ 27,705, matching the six units' sum 6,695+5,203+4,651+3,317+3,312+4,527 (line 21) ✔; army 0 money 186→256, army 2 110→40, Δ 70 (lines 18-19) ✔; `stepperIncrements` `[10, 100]` (line 32) ✔; `autoDisbandOnEmptyArmy` (line 33) ✔. Against `field-recruitment-uniform-attrition-and-fleet-drift.md`: the band `2.50% to 2.85%` is the true max and min of all twelve rows (I checked all twelve: 2.66/2.59/2.85/2.80/2.74/2.58/2.74/2.56/2.82/2.77/2.59/2.50) ✔, the three sample rows in the note are exact ✔, and the note's quotation of the report's "What this does not establish" line is **verbatim**, which I checked specifically because a fabricated quote inside a provenance note would be worse than the original omission ✔. `derived` is the right tag — the report leaves open whether the loss is combat, supply attrition, or both.
  - One tagging call worth stating since I scrutinised it: `armyTransfer.autoDisbandOnEmptyArmy` is `confirmed` while its report lists the branch under "What this does not establish" as *"not exercised in this example, only read from code."* I accept `confirmed` here. The hedge is about the branch not being *observed firing*, not about what the code says — the report decompiled it and names the routine (`FUN_0044ab90`). That matches the convention this corpus applies consistently to code-read-but-unobserved rules (`capture.garrisonClearedOnTransfer`, `diplomacy.allianceContagion`, and ~40 others), and it is categorically different from `victory.yearLimitBC`, where the doubt was about the *meaning of the constant itself*. Crucially, the note carries the caveat verbatim, so a later task can't be misled.
- **P10 — counts verified independently, not accepted.** I regenerated them myself: **355 entries, 347 `confirmed` / 8 `derived` / 0 `designed`**, matching the body's script output exactly. The 8 `derived` ids I enumerated match the body's list one-for-one.
- **`terrain.daTOffset` → `terrain.datOffset`** ✔, value/source/tag unchanged.
- **Duplicate-fact pairs — no value was lost in the replacement.** `caps.maxMercenarySlots` still holds `50` (`mercenary-pool-record.md`) and `caps.meleeLossAbsoluteCap` still holds `30000` (`decompiled-combat-formula-structure.md`); the dropped ids were the redundant copies, and the surviving numbers are the ones `required-ids.json` already guarded. Nothing downstream referenced the dropped ids ✔.

---

## Gate 3 — Determinism. **PASS**

Unchanged and re-checked. No `Random`/`DateTime`/`Guid.NewGuid`/`Stopwatch`/`Environment.*`/`Process.*` anywhere in the Fixtures folder. `FixtureCorpus.All` still preserves `corpus.json`'s file order; hash sets are still used only for `Contains`/`TryGetValue`. `BuildIndex` now throwing on a duplicate is *more* deterministic than last-wins, not less — the failure is the same on every run and every machine.

## Gate 4 — Scope. **PASS**

8 files, all inside `tests/fixtures/**` and `tests/IC2.Engine.Tests/Fixtures/**`. **Zero `docs/` diff** (`git diff main...HEAD --name-status -- docs/` empty). `tests/IC2.Engine.Tests/IC2.Engine.Tests.csproj` untouched, confirmed by path-scoped diff. `Fixtures.cs` → `FixtureCorpus.cs` is a rename within the Owns folder. `DownstreamNamespaceUsageTests.cs` is physically inside the Owns folder. CI-safe: verified by running the exact `ci.yml` Release `--no-build` sequence.

## Gate 5 — Correctness. **PASS**

**C1 is fixed and, more importantly, guarded.** `DownstreamNamespaceUsageTests` declares a sibling namespace, does `using IC2.Engine.Tests.Fixtures;`, and calls `FixtureCorpus.Get(...)` unqualified — the exact shape that produced CS0234 before. I confirmed it is a real guard and not a decorative test: I temporarily edited a copy of that call back to `Fixtures.Get(...)` and got `error CS0234` again, so a regression to the colliding name fails the build in T04's own Owns list rather than being rediscovered by ~14 later tasks. That is the right shape of fix.

The three LOW fixes are all in and verified (`IsNullOrWhiteSpace` for string values, id-non-empty check with an `<empty id>` placeholder in the failure message, `ParseEntry` wrapping malformed entries with array index + id). The two documented-not-changed calls are sound: `JsonElement` record equality is genuinely unused today, and `FindRepositoryRoot`'s duplication has no shared location to hoist to while T02 and T04 own separate folders — which is what I said in round 0.

### Ruling 1 — the `IC2.Engine.Tests.Calendar` namespace: **not a finding.**

I tested this rather than reasoning about it, both directions:

1. **No collision.** I created `tests/IC2.Engine.Tests/Calendar/ZzT06Probe.cs` declaring `namespace IC2.Engine.Tests.Calendar` — exactly what T06 will add, same namespace, its own folder — and the solution built clean, 0 warnings. C# namespaces are open across files and folders, so T06 can own that namespace normally with this file present. The only way to collide would be T06 declaring a type also named `DownstreamNamespaceUsageTests`.
2. **A neutral namespace would guard identically.** I probed `namespace IC2.Engine.Tests.DownstreamUsageProbe` (a sibling of `...Tests.Fixtures` that no task's Owns list will ever claim): it compiles today, and renaming the type back to `Fixtures` reproduces the same `CS0234`. So the guard's power comes from being *a* sibling, not from being `Calendar`.

Since the guard works either way and there is no compile-level cost, this is not a gate failure. The only real cost is cosmetic: xunit reports the test as `IC2.Engine.Tests.Calendar.DownstreamNamespaceUsageTests`, so it will show up under T06's namespace in test output permanently — I saw exactly that in my injected-defect run above. **Non-blocking suggestion, take it or leave it:** if this is ever touched again, `IC2.Engine.Tests.DownstreamUsageProbe` buys the same guarantee and squats on nothing. Not worth a commit on its own, and explicitly not a reason to hold this PR.

### Ruling 2 — the declined consistency test: **the implementer's argument is correct; accepted.**

My round-0 complaint was about *one fact under two ids that can drift*. The three pairs it dropped had **identical value and identical source** — literal copies where an edit to one silently staled the other and nothing identified which was authoritative. `error.fleetCannotAttackDocked` (the literal string) and `battle.naval.dockedFleetCannotBeAttacked` (a boolean rule flag) are not copies: they are a rule and its user-facing message, carrying different information in different shapes. Dropping either loses something a downstream task needs, and drift is not really possible because neither is derivable from the other by copying. It also matches a pattern already established across this corpus that I approved in round 0 — `caps.maxTroopsPerArmy` (100000) alongside `error.armyTooLarge100k` (the string), `recruitment.mobilizationCapPercent` (100) alongside `error.mobilizationAlready100`. And the implementer drew the line in exactly the right place when it *did* drop `recruitment.armyCap100k.errorString`, which was an identical-string-and-source copy of `error.armyTooLarge100k` while `caps.maxTroopsPerArmy` kept the number. Consistent reasoning; no finding.

### One non-blocking observation on the fix shape

The `.evidence` id pattern (`caps.maxMercenarySlots.evidence`, `battle.melee.rawLossAbsoluteCap.evidence`) does solve the drift problem I raised — the number now lives in exactly one place. But it puts prose into a `value` field while leaving the surviving entries' `note` empty, so `caps.maxMercenarySlots` now reads as a bare unexplained `50` unless you happen to look up a sibling id. `FixtureEntry.Note` is the field designed for that text. Folding the two `.evidence` values into their partners' `note` and dropping the two ids would be strictly tidier. **This is a taste call on a fix that achieves its stated goal, not a defect, and explicitly not grounds for another round** — bank it for whenever the corpus is next touched.

---

## Bottom line

| Gate | Round 0 | Round 1 |
| --- | --- | --- |
| 1 DoD | PASS | **PASS** |
| 2 Provenance | FAIL | **PASS** |
| 3 Determinism | PASS | **PASS** |
| 4 Scope | PASS | **PASS** |
| 5 Correctness | FAIL | **PASS** |

Across both rounds: **355 entries, 214+ numeric and 115+ string values, and between my two passes and the independent `/code-review` pass, not one wrong value has been found in any of them.** The rework fixed every finding on its merits rather than on the letter — P1 in particular was resolved by getting the semantics right (two distinct claims, precisely stated) rather than by picking the citation that made the test pass. Approving.

Issue #37, opened 2026-09-13
T30 Harden IC2.Data: army tombstones, and the DAT's own file layout


diegoami:
Spec: [`docs/build-orchestration-plan.md` §5 → T30](https://github.com/diegoami/imperial_conquest_2/blob/main/docs/build-orchestration-plan.md#t30-harden-ic2data-against-tombstone-and-sentinel-records) · index row in §12.

**Branch** `task/T30-data-tombstone-records` · **Model/effort** Haiku / Medium · **Reviewer** Sonnet / Medium
**Start after** T01 · **Merge after** T01 — and **before T21 starts**.

## Why this exists

Closes a backlog gap rather than a design milestone. T01 explicitly rules the existing parsers out of scope ("do not 'fix' existing `IC2.Data`/`IC2.Inspect` code"), and no other task owns `src/IC2.Data/**` — yet **T21 (original-save import) and T29 (classical world export) are both built directly on them**.

## The defect

`SaveArmyTable.Parse` rejects any army record whose owner word exceeds 15 and throws `InvalidDataException`, aborting the parse of the **entire save**. One bad record makes a whole file unreadable to every tool built on the parser.

Real saves contain such records legitimately. `0xFFFF` is the established **no-owner sentinel** — the same one `galatia-elimination-and-city-resupply-confirmed.md` already had to special-case for the capital field — marking an army slot merged or eliminated during the turn and not yet compacted. **Two of the thirteen saves in one ordinary 13-turn series are unreadable for this reason.**

Found by `docs/investigations/thracia-supply-morale.md`, which hit it on real saves and had to decode those two from raw bytes.

## Owns

`src/IC2.Data/**`, `src/IC2.Inspect/**`, `tests/IC2.Data.Tests/**` — the test project T01 already pre-declares in `IC2.sln`, so this task edits no solution file.

## Done when

1. `1_thracia_271_spring_3.sav` and `1_thracia_271_autumn_1.sav` both parse; `--list-armies … Thracia` prints the Thracian army for each instead of aborting.
2. Tombstoned records are **excluded from `Armies`** (not returned as degenerate entries with owner 65535) and exposed separately — e.g. a `SkippedRecords` count — so a caller can tell "clean parse" from "parse with tombstones" without re-reading bytes.
3. A record malformed for any *other* reason still throws, with a message naming the record index and the failing field. Tombstone tolerance must not become blanket tolerance for corruption.
[...]

Issue #32, opened 2026-09-13
T29 Export the shipped classical-mediterranean world and ruleset


diegoami:
**Spec:** [`docs/task-catalogue.md` → T29](https://github.com/diegoami/imperial_conquest_2/blob/main/docs/task-catalogue.md#t29-export-the-shipped-classical-mediterranean-world-and-ruleset). The catalogue entry is the contract: scope, Owns, Done when, hazards, model, reviewer and dependencies. This issue doesn't copy it, so nothing here goes stale.

**Branch:** `task/T29-export-classical-world`

This issue's `status:*` label is its status. How a task runs: [build-process.md §4](https://github.com/diegoami/imperial_conquest_2/blob/main/docs/build-process.md#4-the-task-loop). Earlier revisions of this body copied the catalogue entry; see the edit history.

diegoami:
**Proposed change on branch [`plan/post-thracia-scheduling`](https://github.com/diegoami/imperial_conquest_2/compare/main...plan/post-thracia-scheduling)** — not yet merged, held for review.

**DoD line 1 is currently unsatisfiable.** It requires the export to be "cross-checked against `IC2.Data`'s own parse of the same file (same city count, same nation names, same map dimensions)". As of today, `IC2.Data` cannot parse the DAT at all beyond `WorldPrefix`: **six of its seven parsers throw on it**, because the DAT has no record-count words and its nation record is a different size from the SAV's (see #37, and `docs/investigations/dat-file-layout.md` on the branch).

The good news is that the check can be made *real* rather than dropped — **the DAT does contain a full 16-record nation table** at `0x1B100`, names in `NationCatalog` order, with capital-city index, city count, treasury, unity, mobilized % and tax rate all present at known offsets. So:

- **`merge-after` and `start-after` gain T30**, which delivers the DAT parse this line depends on. T30 moves to `phase:0`/wave 1 to land first; this task stays in wave 2 alongside T03.
- **DoD 1 rewritten** to name what is cross-checked against what: map dimensions and city count from `WorldPrefix`, the 16 nation names *in order* from T30's DAT nation-table parse, and the per-nation fields from the same parse.
- **New DoD 2**: leader names and the human-player flag are **not** exported from the DAT, because they are not in it — `TPremierForm_NewGame` assigns both at New Game (the leader by a 1-in-12 draw from a name pool at `0x2089A`). `NationCatalog` stays a *cross-check* on the DAT parse, never the export's source of record.
[...]

Issue #14, opened 2026-09-13
T14 Naval


diegoami:
**Spec:** [`docs/task-catalogue.md` → T14](https://github.com/diegoami/imperial_conquest_2/blob/main/docs/task-catalogue.md#t14-naval). The catalogue entry is the contract: scope, Owns, Done when, hazards, model, reviewer and dependencies. This issue doesn't copy it, so nothing here goes stale.

**Branch:** `task/T14-naval`

This issue's `status:*` label is its status. How a task runs: [build-process.md §4](https://github.com/diegoami/imperial_conquest_2/blob/main/docs/build-process.md#4-the-task-loop). Earlier revisions of this body copied the catalogue entry; see the edit history.

diegoami:
**Proposed new scope + DoD lines on branch [`plan/post-thracia-scheduling`](https://github.com/diegoami/imperial_conquest_2/compare/main...plan/post-thracia-scheduling)** — not yet merged, held for review.

This task did not mention **fleet storm attrition or the zero-supply penalty at all**, and no other task owned them. Both are `[confirmed]` from code *and* empirically, on the ten-save `1_cartago_271_*` series (`docs/investigations/thracia-supply-morale.md` §"Fleets", `design-audit.md` §2.9a).

Added to **Scope**: the per-turn at-sea attrition pass, registered into the attrition phase T06 declares. The **order is part of the specification**, because two consequences depend on it:

1. Supplies `−= ships`, every turn, every fleet — **in port as well as at sea**.
2. Everything below is at-sea only (`FleetRecord +10 == 0xFFFF`); a fleet in port is exempt.
3. **Storm pass, unconditional and not supply-driven**: `dmg = max(1, random(100 − condition) / 10)`, doubled/capped-5 in Winter, tripled/capped-8 on the `+24 == 1` branch, then `× 2 + 1` away from friendly coast (1-in-20 Winter spike to 30) or halved beside it. `dmg ≥ 6` costs ships too. Because the roll scales with damage already taken this is a **death spiral, not a slope** — and it dominates the supply rider by an order of magnitude.
4. **Death check**: condition `< 40` destroys the fleet with *"A fleet belonging to X is lost at sea."*
5. Moves `= 30 − (ships − 50) / 10`, minus the carried-army term.
6. **Zero supply** (`== 0` exactly, absolute not a percentage): moves −3, condition `−random(0..1)`.
7. Condition `< 70` costs a further `(70 − condition) >> 2` moves.

[...]

Issue #8, opened 2026-09-13
T08 Economy, supply, and purses


diegoami:
**Design milestone:** M3 · **Branch:** `task/T08-economy`
**Model / effort:** Sonnet / High · **Reviewer:** **Opus / Medium**
**Owns:** `src/IC2.Engine/Economy/**`, `tests/IC2.Engine.Tests/Economy/**`, `src/IC2.Engine/Model/Ruleset.cs` (the `EconomyRules` record and new records nested in it only — additive), `data/rulesets/toy-ruleset.json` (the `economy` block only), `tests/fixtures/**` (DoD 13's top-up only, under [build-process.md §2.4](https://github.com/diegoami/imperial_conquest_2/blob/main/docs/build-process.md#2-how-the-build-avoids-conflicts)'s contract)

### Scope
Tax; quarterly upkeep with real non-payment consequences; tribute growth; loyalty drift and the rebellion check's *confirmed structure* with `_provenance`-tagged placeholder thresholds; the weather-event frequency curve with data-driven effects; **supply as a purchased economy**; and **per-army / per-fleet money purses**.

### Done when
- [ ] `income = 2440 × 15 / 100` and `× 20 / 100` reproduce both published Rome figures exactly
- [ ] Ship upkeep = `3 × ships` per quarter
- [ ] The 13-unit Roman roster's regular upkeep computes to exactly **442**
- [ ] That army's 482 tons against 48,173 troops reads exactly **100%**; the supply triple 204/998, 344/998, 184/282 reads **20% / 34% / 65%**
- [ ] A 100-ton supply purchase costs **20** talents, debited from the buying army's purse and credited to the **selling city's owner's** treasury, behind ruleset flag `supplyPurchaseCostsMoney` (default `true`, per the code) whose `_provenance` names `design-audit.md` Q9
- [ ] The purse cap of 1,000 is enforced on every path that credits a purse
- [ ] An army whose upkeep cannot be paid loses troops — a real consequence, not a debt counter
- [ ] Weather events fire ~8× more often in Winter than Summer over a fixed-seed 400-quarter run (a ratio band — the only band assertion in the plan, because the underlying figure is itself approximate)

### Blocked by
- [ ] #3 T03 Engine seams
- [ ] #4 T04 Fixtures corpus
- [ ] #6 T06 Calendar and turn sequencing
- [ ] #60 T32 Make T06's calendar tests independent of later systems

### Note
[...]

PR #38, opened 2026-09-13
Rescope T30 around the DAT file layout, and give the confirmed supply mechanics owners

diegoami:
Doc-only. Settles why some original files can't be opened, and fixes the task order and definitions that finding invalidates. Held for review rather than merged, per the same handling as the last investigation.

## The finding

**The DAT contains a full 16-record nation table** — file offset `0x1B100`, stride **1,055** bytes, names in exact `NationCatalog` order. That was the open question deciding whether T30 is a one-line sentinel fix or real parsing work. It is real parsing work.

Settled by decompiling **`FUN_004481a0`** (`0x004481A0`), the DAT loader, reached as an indirect call from `TPremierForm_NewGame` (`0x0045a9e0`) — which is why it never appeared in the whole-application dump and was never traced. Its read sequence **accounts for all 140,706 bytes of the file exactly**, and lands the season table on `0x1F7D8`, independently corroborating the offset `docs/investigations/thracia-supply-morale.md` solved from raw bytes. Its army and fleet array bases are the same two globals the turn tick walks, and the DAT's army 14 and fleet 0 are the exact starting states of that investigation's Thracian army and Carthaginian fleet.

Three consequences:

- **No count words anywhere.** The loader reads 15 army and 2 fleet records, then *assigns* `DAT_004a0324 = 15` / `DAT_004a0326 = 2`. `SaveNationLayout.Locate` reads army 0's `x` (100) as an army count.
- **The nation record is a different record** — 1,055 vs. the SAV's 1,172, missing 117 bytes in **five separate places**, not shifted by a constant. No adjustment to `Locate` can reach it.
- **Leader names and the human-player flag are genuinely not in the DAT.** `FUN_00448aa4` assigns both at New Game. T29 cannot source them from the file because the original doesn't either.

## Failure accounting — every file, every parser

51 `.sav` files plus the DAT, through all seven `IC2.Data` parsers:

| Class | Count | Fails | Cause |
| --- | ---: | --- | --- |
| Clean | **48 saves** | — | all seven parsers succeed |
| Army tombstone | **3 saves** | `SaveArmyTable` only | one `owner == 0xFFFF` record aborts the file |
| DAT layout | **1 DAT** | **6 of 7** | no count words; different nation record |

`1_thracia_271_spring_3.sav` (army 10), `1_thracia_271_autumn_1.sav` (army 9) and **`1_cartago_271_spring_5.sav` (army 0) — new**, missed by the earlier Thracia-only sample. One record each, always with **valid coordinates**, so the owner word alone is the discriminator. The previously-reported "3 of 27" was 2 saves plus the DAT out of a 27-file sample.

## What changes

| Task | Change |
| --- | --- |
| **T30** | `phase:2` → **`phase:0`**, into **wave 1** (it was in no wave at all). Haiku/Medium → **Sonnet/High**, reviewer → **Opus/Medium**. Retitled off "tombstone records". Two defect classes. DoD 5 → 8 lines: whole-corpus sweep replacing the sampled check, a DAT nation parse, `Leader`/`HumanPlayer` absent-by-construction, typed rejection for neither-shape files. Now gates **T29 as well as T21**. |
| **T29** | Moved after T30; `start`/`merge-after` gain T30. DoD 1 made satisfiable (cross-check against T30's DAT parse); new DoD 2 for the absent fields; hazard against the `NationCatalog` tautology fallback. |
| **T08** | Owns the supply→morale rule, per-turn consumption and the season table. New DoD 10–13, incl. the Thracian series replayed turn-for-turn and the **47th-report fixtures top-up** (a top-up under T04's existing Owns list, **not** a reopen). |
| **T06** | Owns *when*. New DoD 6–7: attrition before the calendar advance, every turn, every seat — including that it observes the **pre-advance** season, which T08's figures are off by 7× without. |
| **T14** | Owns the at-sea attrition pass (storm spiral, death check, zero-supply penalty), ordering as spec. New DoD 10–14 incl. the Cartago moves series against its supplied control fleet. |
| **T07** | New DoD 6: 51–70 as a consumed invariant. Explicitly does **not** implement the rule. |

Plus: dependency graph gains `T30 → T29`; wave table; a "looked parallel and is not, corrected" entry; reviewer table; §2.4 gains a corpus-currency rule; and the totals line is corrected (it read "29 tasks" and omitted T30 from the model tally).

New file `docs/investigations/dat-file-layout.md`; `design-audit.md` gains §2.14 `[confirmed, new]`.

## Review notes

- **Doc-only — no `src/` changes.** The parser sweep used a throwaway console project outside the repo; its permanent form is T30's DoD line 4.
- Every byte offset is traced to `FUN_004481a0`/`FUN_00448aa4`, not recovered by searching a SAV for matching values — T30's hazard note makes that distinction a merge gate for the implementer too.
- Board state was corrected live and separately (#6/#7/#9/#10/#11 → `status:ready`; #37 gained its missing `task` label). See #29.
- **Dispatch ordering**: T30 should be dispatched against *this* branch's DoD, so this merges first. Plan §6.4 forbids an implementer working to an unmerged DoD.

Closes nothing; updates #37, #32, #8, #14, #6, #7.


diegoami:
Catalogue entry amended by PR #90 (T41 thin CLI demo, and the #87/#88 folds).

PR #39, opened 2026-09-13
T30 Harden IC2.Data: army tombstones, and the DAT's own file layout

### LINKED ISSUE #37: T30 Harden IC2.Data: army tombstones, and the DAT's own file layout

diegoami:
Closes #37

## Task

- **Task**: T30 Harden `IC2.Data`: army tombstones, and the DAT's own file layout — [plan entry](https://github.com/diegoami/imperial_conquest_2/blob/main/docs/build-orchestration-plan.md#t30-harden-ic2data-army-tombstones-and-the-dats-own-file-layout)
- **Branch**: `task/T30-data-parser-hardening`

| --- | --- |
| `src/IC2.Data/SaveArmyTable.cs` | `src/IC2.Data/**` |
| `src/IC2.Data/SaveFleetTable.cs` | `src/IC2.Data/**` |
| `src/IC2.Data/SaveNationTable.cs` | `src/IC2.Data/**` |
| `src/IC2.Data/SaveRecruitmentTable.cs` | `src/IC2.Data/**` |
| `src/IC2.Data/SaveMercenaryTable.cs` | `src/IC2.Data/**` |
| `src/IC2.Data/SaveTurnState.cs` | `src/IC2.Data/**` |
| `src/IC2.Data/SaveFormat.cs` (new) | `src/IC2.Data/**` |
| `src/IC2.Data/DatLayout.cs` (new) | `src/IC2.Data/**` |
| `src/IC2.Inspect/Program.cs` | `src/IC2.Inspect/**` |
| `tests/IC2.Data.Tests/IC2.Data.Tests.csproj` | `tests/IC2.Data.Tests/**` |
| `tests/IC2.Data.Tests/LocalAssets.cs` (new) | `tests/IC2.Data.Tests/**` |
| `tests/IC2.Data.Tests/SyntheticSaveBuilder.cs` (new) | `tests/IC2.Data.Tests/**` |
| `tests/IC2.Data.Tests/ArmyTombstoneTests.cs` (new) | `tests/IC2.Data.Tests/**` |
| `tests/IC2.Data.Tests/DatFormatTests.cs` (new) | `tests/IC2.Data.Tests/**` |
| `tests/IC2.Data.Tests/CorpusFixtures/CorpusEntry.cs` (new) | `tests/IC2.Data.Tests/**` |
| `tests/IC2.Data.Tests/CorpusFixtures/CorpusFixture.cs` (new) | `tests/IC2.Data.Tests/**` |
| `tests/IC2.Data.Tests/CorpusFixtures/CorpusSweepTests.cs` (new) | `tests/IC2.Data.Tests/**` |
| `tests/IC2.Data.Tests/CorpusFixtures/expected-corpus-outcomes.json` (new) | `tests/IC2.Data.Tests/**` |

- [x] No change to `docs/build-orchestration-plan.md`, original game files, `assets.local.ini`, `.gitignore`'s exclusion policy, or anything under `docs/reports/`.

Note on `tests/IC2.Data.Tests/IC2.Data.Tests.csproj`: T01's comment there says later tasks add test files "without editing this project's .csproj". This task is the one justified exception: Done-when line 8 requires an explicit **Skipped** test-runner result (not a silent pass) when `assets.local.ini` is absent, and xunit 2.9.2's plain `[Fact]` has no runtime skip, only a compile-time `Skip` string. Added `Xunit.SkippableFact` (a small, test-only, widely-used xunit v2 extension providing `[SkippableFact]`/`[SkippableTheory]` + `Skip.IfNot`) rather than inventing an ad hoc "print and return early" convention that would just show as an ordinary pass. No production (`IC2.Data`/`IC2.Inspect`) dependency changed.

`tests/IC2.Data.Tests/CorpusFixtures/expected-corpus-outcomes.json` lists the 51 real save **filenames** (already cited individually elsewhere in the committed investigation docs) plus per-file record counts generated by running this branch's own parsers over the configured corpus — no save bytes, screenshots, or other original game files are included.

## DoD evidence

Each of the 8 "Done when" lines from the T30 task entry, in order. All commands were run on this machine with `assets.local.ini` configured (`C:\Users\diego\Documents\imp_conq_original`).

```text
DoD 1: "1_thracia_271_spring_3.sav", "1_thracia_271_autumn_1.sav" and "1_cartago_271_spring_5.sav"
all parse; "--list-armies … Thracia" prints the Thracian army for the first two instead of aborting.

$ dotnet run --project src/IC2.Inspect -- --list-armies ".../saves/1_thracia_271_spring_3.sav" Thracia
Army 6 at (160, 30) · 22,000 troops · 98 tons supply (44%) · 100 money · moves 9 · morale 66

$ dotnet run --project src/IC2.Inspect -- --list-armies ".../saves/1_thracia_271_autumn_1.sav" Thracia
Army 6 at (160, 30) · 22,000 troops · 0 tons supply (0%) · 0 money · moves 8 · morale 51

(both values match docs/investigations/thracia-supply-morale.md's table exactly; both saves
previously aborted the whole parse with InvalidDataException)

DoD 2 + 3: tombstones excluded from Armies and surfaced via SkippedRecords; any other malformed
record still throws, naming the record index and the failing field.

$ dotnet test tests/IC2.Data.Tests/IC2.Data.Tests.csproj --filter "FullyQualifiedName~ArmyTombstoneTests"
Passed IC2.Data.Tests.ArmyTombstoneTests.Known_tombstone_saves_skip_the_record_and_keep_parsing(relativePath: "saves/1_thracia_271_spring_3.sav", expectedIndex: 10, expectedX: 41, expectedY: 62)
Passed IC2.Data.Tests.ArmyTombstoneTests.Known_tombstone_saves_skip_the_record_and_keep_parsing(relativePath: "saves/1_thracia_271_autumn_1.sav", expectedIndex: 9, expectedX: 38, expectedY: 52)
Passed IC2.Data.Tests.ArmyTombstoneTests.Known_tombstone_saves_skip_the_record_and_keep_parsing(relativePath: "saves/1_cartago_271_spring_5.sav", expectedIndex: 0, expectedX: 97, expectedY: 31)
Passed IC2.Data.Tests.ArmyTombstoneTests.Tombstone_is_skipped_even_among_other_valid_and_would_be_invalid_records
Passed IC2.Data.Tests.ArmyTombstoneTests.Owner_code_out_of_range_and_not_the_sentinel_still_throws_naming_index_and_field
Passed IC2.Data.Tests.ArmyTombstoneTests.Out_of_range_x_coordinate_still_throws_naming_index_and_field
Passed IC2.Data.Tests.ArmyTombstoneTests.Out_of_range_y_coordinate_still_throws_naming_index_and_field
Passed IC2.Data.Tests.ArmyTombstoneTests.Unit_slot_out_of_range_type_still_throws_naming_index_slot_and_field
Passed IC2.Data.Tests.ArmyTombstoneTests.Unit_slot_out_of_range_quality_still_throws_naming_index_slot_and_field
Passed IC2.Data.Tests.ArmyTombstoneTests.Unit_slot_missing_name_still_throws_naming_index_slot_and_field
... 12 Passed

DoD 4: a regression fixture runs all 51 saves plus the DAT through all seven parsers against a
committed expected-outcome table.

$ dotnet test tests/IC2.Data.Tests/IC2.Data.Tests.csproj --filter "FullyQualifiedName~CorpusSweepTests"
Passed!  - Failed: 0, Passed: 53, Skipped: 0, Total: 53
(52 file cases [51 saves + the DAT] x the seven-parser assertion, plus 1 corpus-coverage
cross-check confirming the fixture enumerates every .sav currently on disk)

DoD 5: the DAT parses to 16 nation records matching NationCatalog in order, city counts summing to
334, 15 armies, 2 fleets.

$ dotnet test tests/IC2.Data.Tests/IC2.Data.Tests.csproj --filter "FullyQualifiedName~DatFormatTests"
Passed IC2.Data.Tests.DatFormatTests.Dat_yields_16_nations_in_catalog_order_with_city_counts_summing_to_334
Passed IC2.Data.Tests.DatFormatTests.Dat_yields_15_armies_and_2_fleets
(Rome's DAT-parsed treasury/unity/capital/cities read 2200/821/85/25, matching
docs/investigations/dat-file-layout.md's cited cross-check against the SAV path exactly)

DoD 6: Leader and HumanPlayer are absent from a DAT parse by construction, never defaulted; a
negative test proves a caller cannot read a leader name off the DAT and get a non-empty string.

Passed IC2.Data.Tests.DatFormatTests.Leader_is_absent_from_every_dat_nation_not_a_fabricated_empty_string
Passed IC2.Data.Tests.DatFormatTests.HumanPlayer_throws_rather_than_defaulting_on_a_dat_nation
Passed IC2.Data.Tests.DatFormatTests.Sav_nation_record_still_carries_a_real_leader_and_human_player_flag

DoD 7: a file that is neither DAT-shaped nor SAV-shaped is rejected with a typed error naming which
discriminator failed.

Passed IC2.Data.Tests.DatFormatTests.A_file_matching_neither_shape_is_rejected_naming_both_checks
Passed IC2.Data.Tests.DatFormatTests.An_empty_file_is_rejected_not_silently_treated_as_either_shape
Passed IC2.Data.Tests.DatFormatTests.A_file_the_exact_dat_length_but_garbage_content_is_still_detected_as_dat_shaped
Passed IC2.Data.Tests.DatFormatTests.A_real_save_is_detected_as_sav_shaped
Passed IC2.Data.Tests.DatFormatTests.Dat_is_detected_as_dat_shaped

DoD 8: every test in this task skips with an explicit "original files not configured" result when
assets.local.ini is absent.

$ mv assets.local.ini assets.local.ini.bak && dotnet test tests/IC2.Data.Tests/IC2.Data.Tests.csproj
Passed!  - Failed: 0, Passed: 11, Skipped: 65, Total: 76
(the 11 passes are the pure-synthetic tests in ArmyTombstoneTests/SyntheticSaveBuilder/
DatFormatTests's format-discriminator group, which construct their own bytes and need no local
prerequisite; every asset-dependent test reports "Skipped" with reason "original files not
configured", not a silent pass or a failure)
$ mv assets.local.ini.bak assets.local.ini   # restored immediately after

Full solution, config restored:
$ dotnet build IC2.sln
Build succeeded. 0 Warning(s) 0 Error(s)
$ dotnet test IC2.sln
Passed! - IC2.Data.Tests.dll: Failed 0, Passed 76, Skipped 0, Total 76
Passed! - IC2.Engine.Tests.dll: Failed 0, Passed 156, Skipped 0, Total 156
```

## Provenance checklist

- [x] Every constant introduced or changed traces to `docs/investigations/dat-file-layout.md` (DAT offsets: army table 0x18A5C = `WorldPrefix.SharedPrefixLength`, fleet table 0x1B0CC, nation table 0x1B100 / 1,055-byte stride, and the six field offsets cited verbatim in that doc's "Mapped onto the DAT record's own offsets" line — treasury 0x40D, unity 0x411, mobilized 0x413, capital 0x415, cities 0x417, tax 0x419 — plus the DAT's total length 140,706 and its army/fleet counts 15/2) or to `docs/investigations/thracia-supply-morale.md` (the 0xFFFF tombstone sentinel). The one number not directly stated in either doc as a hex offset — the recruitment queue's DAT-local offset 0x2C9 (713) — is the running sum of that same read-order table's preceding row lengths (11 + 32 + 2 + 668), the identical arithmetic the table's own 1,055-byte total is computed with; not recovered by searching a SAV for matching values.
- [x] No `[designed]` value was introduced by this task.

Values omitted because no source was found:

- `NationRecord.Leader` and `NationRecord.HumanPlayer` are omitted (modelled as absent, not defaulted) on a DAT-origin record, exactly because no source exists for them in the DAT — see DoD 6 above.
- `SaveMercenaryTable`/`SaveTurnState` throw `DatDataNotPresentException` on a DAT-shaped file rather than returning an invented empty pool / zeroed calendar, for the same reason: the DAT has no mercenary-offer pool and no calendar/current-turn trailer at all, and no decompiled code assigns either at DAT-load time (only at New Game, via a separate helper — see docs/investigations/dat-file-layout.md).

## Determinism

- [x] This PR touches no gameplay path — `IC2.Data`/`IC2.Inspect` are file-format parsers with no randomness, no wall-clock reads, and no `IRng` involvement.

## Notes for the reviewer

- **Defect A audit**: I checked every other range-check in `IC2.Data` for the same unguarded-sentinel pattern. Only `SaveArmyTable`'s owner check had it. `SaveNationTable` already special-cased `0xFFFF` for the capital-city field (pre-existing code, unchanged). `SaveFleetTable` never validated its owner field at all (unchanged). `WorldPrefix` (city table) never validated an owner field either (unchanged). `SaveTurnState.Parse` has an unguarded `currentNation > 15` check, but neither investigation doc shows evidence of a current-nation tombstone case, and the corpus sweep found no failure there across all 51 saves — so, per the hazard note ("do not widen the owner check to accept all values ≤ 65535"), I left it unchanged rather than guessing at an unevidenced sentinel.
- **`IC2.Inspect --to-json`** (`SaveJsonExporter.cs`) still throws unhandled on the DAT (it always did, via a different exception before this change) — exporting the DAT to JSON is explicitly T29's job per the plan, not T30's, so I left that path alone rather than partially duplicating T29's design. I did fix `--inspect-nation`'s crash on the DAT (`Program.cs`), since that crash was a direct, newly-exposed consequence of this task's own change (DAT nation parsing now succeeds, where it used to abort before reaching the `HumanPlayer` read) and the fix is a few lines within `src/IC2.Inspect/**`.
- **`godot/MapViewer.cs`** is outside every Owns list and outside `IC2.sln` (excluded from the solution entirely — see `Directory.Build.props`). I confirmed it compiles unaffected: it never calls `SaveArmyTable`/`SaveFleetTable`/`SaveNationTable`/`SaveRecruitmentTable`/`SaveTurnState` on DAT bytes (only `WorldPrefix.Parse`, which is format-agnostic and unchanged), and `NationRecord.HumanPlayer` deliberately stayed a non-nullable `bool` (backed by a nullable field that throws when absent) specifically so that file's `nation.HumanPlayer ? … : …` ternaries keep compiling; only `Leader` became nullable (`string?`), which is compile-safe there.
- No evidence conflicts found. The corpus sweep's live numbers (51 saves: 48 clean + 3 with exactly one tombstone each; DAT: 5 of 7 parsers succeed, 2 throw the expected typed absence exception) match `docs/investigations/dat-file-layout.md`'s "The parse sweep" table exactly.


diegoami:
## Review — T30 (#37), PR #39 · Opus / Medium · round 1

**Verdict: request changes.** Gates 1–4 pass and I reproduced all eight DoD lines myself. Gate 5 has **one blocking finding**: the DAT/SAV discriminator misclassifies a genuine SAV, reproducibly, and I have a working repro below. Everything else is a note, not a blocker.

This is a strong PR — the provenance discipline in particular is the best I've seen on this build (see gate 2). The blocking item is a three-line fix in one function.

---

### Gate 1 — DoD, reproduced by me

Checked out at PR head `5de4da8`, `assets.local.ini` configured. I ran every line myself rather than reading the PR body.

| # | Done-when | Result |
| --- | --- | --- |
| 1 | The 3 tombstone saves parse; `--list-armies … Thracia` prints the army | **Pass.** Ran the CLI directly: `spring_3` → `Army 6 at (160, 30) · 22,000 troops · 98 tons supply (44%) · morale 66`; `autumn_1` → `… 0 tons supply (0%) · morale 51`. Both match `thracia-supply-morale.md`'s table (lines 146, 157) exactly. `1_cartago_271_spring_5.sav` also parses. |
| 2 | Tombstones excluded from `Armies`, surfaced separately | **Pass.** `SaveArmyTable.SkippedRecords` (`src/IC2.Data/SaveArmyTable.cs:29`) is a separate list; `ArmyTombstoneTests.cs:37-38` asserts no owner-65535 entry and no entry at the tombstone index survives in `Armies`. |
| 3 | Any *other* malformed record still throws, naming index and field | **Pass.** The single combined message was correctly split into five distinct ones (`SaveArmyTable.cs:83,85,87,100,102,104`), each naming the index and the offending field *and value*. Six negative tests cover X, Y, owner, unit type, unit quality, unit name. |
| 4 | 51-save + DAT corpus sweep through all seven parsers vs a committed table | **Pass — and it genuinely covers what it claims** (see below). |
| 5 | DAT → 16 nations in `NationCatalog` order, cities summing to 334, 15 armies, 2 fleets | **Pass**, and I cross-checked it independently (see gate 2). |
| 6 | `Leader`/`HumanPlayer` absent by construction, never defaulted | **Pass — I checked specifically for a disguised default.** There is none (see below). |
| 7 | Neither-shape file rejected with a typed error naming which discriminator failed | **Pass as written**, but see the blocking finding — the *converse* case is broken. |
| 8 | Every test skips explicitly when `assets.local.ini` is absent | **Pass.** I renamed the file and re-ran: `Failed: 0, Passed: 11, Skipped: 65, Total: 76`, skip reason `original files not configured`. The 11 passes are genuinely synthetic tests (`SyntheticSaveBuilder`-built bytes + the format-discriminator group) plus T01's pre-existing placeholder — not asset tests silently passing. Config restored. |

Full suite at PR head: `IC2.Data.Tests 76/76`, `IC2.Engine.Tests 156/156`, `dotnet build IC2.sln` clean, 0 warnings.

**DoD 4, verified rather than taken on trust.** I did not just check that a test named "corpus sweep" exists:

- The theory enumerates **52 real cases** — I read the per-case test output, 51 `.sav` + `Imperial Conquest 2.dat`, not a single collapsed case.
- `find` over the assets directory returns exactly 51 `.sav` files, matching the fixture's row count.
- `CorpusSweepTests.cs:82-106` re-enumerates the three save directories from disk at run time and asserts set equality with the committed table, so the fixture cannot go stale the way the earlier sampled check did. That is the right shape for this DoD line.
- Each case really does exercise **all seven** parsers (`WorldPrefix`, `SaveArmyTable`, `SaveFleetTable`, `SaveNationTable`, `SaveRecruitmentTable`, `SaveMercenaryTable`, `SaveTurnState`) — `CorpusSweepTests.cs:31-71` — with per-file city/army/skipped/fleet/nation/recruitment/mercenary/week counts, not just "did not throw".
- The fixture exercises real variation: fleet counts 2–6, army counts 8–17, three files carrying exactly one tombstone each — the same three `dat-file-layout.md` names.

One bounded caveat, not a blocker: the expected values were generated by running this branch's own parsers, so the table is a *regression* snapshot, not an independent oracle. That is what the DoD line asks for, and the circularity is fenced in by the invariants asserted independently (334 cities on every row, 16 nations, sum-to-334, the 3-tombstone count, the file-set cross-check). Worth stating explicitly in the fixture's doc comment.

Incidental observation from the fixture, for whoever picks up T29: three saves (`1_rome_270_winter_7`, `1_rome_270_winter_9`, `1_thracia_271_autumn_1`) have nation-declared city counts summing to **339**, not 334. That is an original-game data quirk faithfully recorded, not a parser defect — flagging it only because T29's DoD line 1 asserts 334, and it must take that from `WorldPrefix`, never from the nation records.

**DoD 6, checked specifically for a disguised default.** It is honest:

- `Leader` became `string?` and is set to literal `null` on the DAT path (`SaveNationTable.cs:63`, `leader: null`) — **not** `""`, not `"Unknown"`.
- `HumanPlayer` is backed by a `bool? _humanPlayer` and the public getter **throws** `DatDataNotPresentException` when it is null (`SaveNationTable.cs:135`) — **not** `?? false`. I verified the negative test actually asserts the throw for all 16 nations (`DatFormatTests.cs:97-98`), and that the SAV path still returns real values (`DatFormatTests.cs:111-117`).
- `NationRecord.Source` is the explicit "came from the DAT" marker the DoD line asks for.
- The same discipline is applied one level up: `SaveMercenaryTable`/`SaveTurnState` throw `DatDataNotPresentException` on the DAT rather than returning an invented empty pool or a zeroed calendar, and the corpus fixture models those two columns as `null` rather than `0` (`CorpusEntry.cs:24-25`). That is the right call and it is documented.

Keeping `HumanPlayer` a non-nullable `bool` (so `godot/MapViewer.cs`'s ternaries still compile) while backing it with a nullable field is a neat way to satisfy the DoD without reaching outside Owns. Noted and endorsed.

---

### Gate 2 — Provenance: **pass**, and unusually well done

Every DAT constant in the diff traces to `docs/investigations/dat-file-layout.md`, and none of it was recovered by byte-searching a SAV — which is the specific hazard the task entry calls out. I checked this rather than accepting the PR's checklist.

- Verbatim from the doc's read-order table: `0x18A5C` (= `WorldPrefix.SharedPrefixLength`, 100,956), `0x1B0CC`, `0x1B100`, `1055`, `15`, `2`, `656`, `26`, `140706`, name length `11`.
- Verbatim from the doc's "Mapped onto the DAT record's own offsets" line: `0x40D`, `0x411`, `0x413`, `0x415`, `0x417`, `0x419`.
- I re-derived the whole chain myself from the doc's in-memory read-order table and it closes exactly: `11 + 32 + 2 + 668 = 713 = 0x2C9`, `+320` (recruitment) `= 1033`, `+4` (the unlabelled `+0x430` read) `= 1037 = 0x40D` — landing on the doc's cited treasury offset — and running on through unity/mobilized/capital/cities/tax to `1055`, the doc's stated stride. The one constant not spelled out in hex (`0x2C9`) is therefore arithmetic on cited lengths that is *self-checking against a separately cited offset*, not a byte-search. The implementer disclosed it as such in both the PR body and `DatLayout.cs:34-39`. This is exactly the right handling.
- No scanning constructs were introduced anywhere in the diff (no `IndexOf`/search over `data` for magic values beyond the pre-existing NUL-terminator scan).
- No `[designed]` tag was introduced, so `design-audit.md` §4.5's rule does not bite here.

I then confirmed the offsets empirically by a route the implementer did not use — comparing the DAT parse against the **SAV** parse of the game's first save, which reaches the same fields through a completely different offset table:

- `--inspect-nation` on the DAT vs `saves-processed/processed/1.sav` agrees on **name, capital, city count, tax and unity for all nations checked**; Rome reads treasury 2200 / unity 821 / capital 85 / cities 25, matching the doc's cited cross-check exactly. The only divergences (Seleucid, Ptolemaic, Gaul treasury and mobilized) are precisely the fields the AI has already moved by the time save 1 was written.
- The derived `0x2C9` recruitment offset is confirmed the same way: `--inspect-city … Rome` returns **byte-identical garrison contents from the DAT and from save 1** (light infantry 5,500 / heavy infantry 3,500 / heavy infantry 4,000 / heavy cavalry 1,000, same state codes). Reading 55 well-formed entries that all pass the `type > 4 || cityIndex >= 334` validation is itself near-conclusive; matching the SAV path exactly settles it.

**Defect A audit (DoD's "audit the other tables") — verified.** I checked every `InvalidDataException` site in `src/IC2.Data/` myself. The PR's claims hold: only `SaveArmyTable`'s owner check had the unguarded-sentinel pattern; `SaveNationTable:43` already special-cased `0xFFFF` for the capital field; `SaveFleetTable` and `WorldPrefix` never validated an owner at all; `SaveTurnState:48`'s `currentNation > 15` is unguarded but no evidence of a current-nation tombstone exists and the sweep found no failure there. Leaving it alone rather than guessing at an unevidenced sentinel is the correct reading of the hazard note. Consider moving that audit out of the PR body into a comment so it survives the merge.

---

### The `0xFFFF`-only hazard — verified in the code, not from the description

The task entry flags this twice. The implementation is correct and **not** widened:

```csharp
// SaveArmyTable.cs:78-88
if (owner == ArmyRecord.TombstoneOwnerSentinel) { skipped.Add(...); continue; }
if (x >= WorldPrefix.MapWidth)  throw ...;
if (y >= WorldPrefix.MapHeight) throw ...;
if (owner > 15)                 throw ...;
```

Exact equality against `0xFFFF`, nothing else; every other out-of-range owner still throws — no `<= 65535`, no range, no `>= 0xFFF0`, no "any large value". `TombstoneOwnerSentinel` is deliberately kept as a constant distinct from `AboardFleetSentinel` despite the identical bit pattern (`SaveArmyTable.cs:151-155`), which is the right call. `ArmyTombstoneTests.cs:106-118` pins it with owner `999`.

Two small suggestions, neither blocking:

1. Add a boundary case at `owner: 16` alongside `999`. `16` is the value most likely to be let through by a future "off-by-one" relaxation; `999` is not.
2. `SaveArmyTable.cs:78-82` skips a tombstone **before** validating its coordinates, so a tombstone with corrupt coordinates is silently swallowed. The evidence says tombstones always carry valid coordinates (`dat-file-layout.md:154-167`), which makes coordinates a usable corroborating check — asserting them rather than ignoring them would keep DoD 3's "tombstone tolerance must not become blanket tolerance" tighter. The current behaviour is defensible and documented; I am flagging the choice, not calling it wrong.

---

### Gate 3 — Determinism: **pass**

No `System.Random`, `new Random`, `DateTime`/`DateTimeOffset`, `Guid.NewGuid`, `Stopwatch`, or `Environment.TickCount` anywhere in the diff. No `IRng` involvement needed — as expected for a parser, and no gameplay path is touched. The only ordering-sensitive code is `CorpusSweepTests.cs:92,97`, where `Directory.GetFiles`' OS-dependent order is normalised by an explicit `OrderBy(p => p, StringComparer.Ordinal)` on **both** sides before comparison. That is correct and deliberate. No dictionary/set iteration order is relied on.

---

### Gate 4 — Scope: **pass**

All 18 changed files fall inside T30's Owns list (`src/IC2.Data/**`, `src/IC2.Inspect/**`, `tests/IC2.Data.Tests/**`). I verified by diffing the file list against those three prefixes: **zero** files outside. In particular **no diff to `docs/build-orchestration-plan.md`**, and nothing under `docs/reports/`, no original game files, no `.gitignore` policy change.

- **`src/IC2.Inspect/Program.cs` is in scope and is not creep.** `IC2.Inspect/**` is explicitly in Owns. The change is 5 lines and is a direct consequence of this task's own work: DAT nation parsing now *succeeds*, so `--inspect-nation` reaches the `HumanPlayer` read that previously was never hit. I ran it — DAT prints `Rome · not stored in the DAT · leader (not stored in the DAT) · capital Rome`, SAV still prints `computer player · leader Publius Scipio`. Correctly minimal: it did **not** touch `SaveJsonExporter`/`--to-json`, which is T29's job, and the PR says so.
- **`tests/IC2.Data.Tests/IC2.Data.Tests.csproj`** is inside Owns. T01's in-file comment ("later tasks add test files without editing this .csproj") is a convention, not the Owns list, and DoD 8 genuinely cannot be met with xunit 2.9.2's compile-time-only `Skip`. Adding `Xunit.SkippableFact` is the right answer and it is test-only — no production dependency changed. The justification is written into the csproj comment itself, which is the correct place for it. Accepted.
- `expected-corpus-outcomes.json` contains only filenames (already cited individually in committed docs) and derived counts — no save bytes. Correct.

---

### Gate 5 — Correctness: **FAIL (one blocking finding)**

Method: a high-effort sweep over `main...HEAD` plus my own read of every changed file. Every finding below I reproduced myself at PR head.

#### BLOCKING — `src/IC2.Data/SaveFormat.cs:45` — a valid SAV of exactly 140,706 bytes is misclassified as a DAT

`Detect` decides "DAT" on **file length alone**, and checks it **before** the SAV walk:

```csharp
if (data.Length == DatFileLength) return SaveFileFormat.Dat;   // SaveFormat.cs:45
```

Any file of exactly 140,706 bytes is classified `Dat` regardless of content — including a perfectly well-formed SAV. Every one of the six parsers then reads it at DAT offsets. This is the "silently fall through to the wrong layout" case DoD 7 exists to prevent, approached from the other side.

This is **not** hypothetical. The largest save in the corpus is 134,065 bytes — only **6,641 bytes** below the DAT length, i.e. about ten more army records. Repro, at PR head:

```
# take a real save and extend it to exactly the DAT length
$b = [IO.File]::ReadAllBytes(".../saves/1_cartago_271_spring_1.sav")   # 131,313 bytes, parses clean
$pad = New-Object byte[] 140706; [Array]::Copy($b,0,$pad,0,$b.Length)
[IO.File]::WriteAllBytes("$env:TEMP\t30_padded_sav.sav", $pad)

$ dotnet run --project src/IC2.Inspect -- --inspect-nation "$env:TEMP\t30_padded_sav.sav" Rome
Unhandled exception. System.IO.InvalidDataException: Missing nation text at 1B100.
   at IC2.Data.SaveNationTable.ParseDat(...) SaveNationTable.cs:line 63

$ dotnet run --project src/IC2.Inspect -- --list-armies "$env:TEMP\t30_padded_sav.sav" Rome
Unhandled exception. System.IO.InvalidDataException: Army 0 has out-of-range owner code 33.
   at IC2.Data.SaveArmyTable.Parse(...) SaveArmyTable.cs:line 87
```

Note what `owner code 33` is: it is the real army 0's **Y coordinate** (the unpadded file reads `Army 0 at (97, 33)`), read two bytes early because the DAT path skips the SAV's count word. A readable save becomes unreadable, and the diagnostic points at the wrong field — the same failure shape T30 was created to eliminate, just with the formats swapped.

`DatFormatTests.cs:141-150` documents this ("garbage content of the right length is still classified as Dat by `Detect()` itself"), which is honest, but the reasoning there only considers *garbage* of that length. A valid SAV of that length is the case that matters, and it is the one the plan's own wording rules out.

**Suggested fix — three lines, entirely inside Owns, entirely evidence-backed.** Make the DAT arm content-confirming instead of length-only, using a fact `dat-file-layout.md:120-124` already states (*"the 16 names at stride 1,055 read Rome, Carthage, … in `NationCatalog`'s exact order"*) and that `ParseDat` already relies on:

```csharp
if (data.Length == DatFileLength && LooksLikeDatNationTable(data))
    return SaveFileFormat.Dat;
```

where `LooksLikeDatNationTable` checks that record 0's 11-byte name at `DatLayout.NationTableStart` reads `NationCatalog.Name(0)`. A real DAT still matches; a 140,706-byte SAV then falls through to the SAV walk and parses correctly; a 140,706-byte garbage file fails **both** arms and gets the `UnrecognizedSaveFormatException` naming both, which is what DoD 7 asks for and is strictly better than today's behaviour. Please add a test: a SAV-shaped byte array built at exactly `DatFileLength` must `Detect` as `Sav`. `SyntheticSaveBuilder.MinimalSav` can produce one by choosing the army count that lands on that length, or by padding.

#### Non-blocking notes

1. **`SaveFormat.cs:67` — `UnrecognizedSaveFormatException` is not an `InvalidDataException`, and that changes a caller's error contract.** Because `Detect` now runs the `Locate` walk at the top of six parsers, a malformed SAV that previously threw `InvalidDataException` now throws `UnrecognizedSaveFormatException`. `godot/MapViewer.cs:389-393` has three `catch (InvalidDataException)` blocks providing graceful degradation for exactly that case; they will no longer catch it, and the viewer will crash instead of printing "Nation details unavailable". `MapViewer.cs` is outside T30's Owns so it cannot be fixed here — flagging it for the orchestrator to route to whoever owns the Godot lane. (`InvalidDataException` being sealed is correctly noted in the code, so the exception hierarchy is not the implementer's fault.) Net effect on that file is still positive: `MapViewer.cs:382` calls `SaveArmyTable.Parse` *outside* any try, so the three tombstone saves used to take the whole viewer down and now do not.

2. **`SaveArmyTable.cs:51-56` and `SaveFleetTable.cs:34-48` — the SAV-branch structural guards are now unreachable.** `Detect` runs `SaveNationLayout.Locate` first, and `Locate` already rejects everything those guards reject (short-before-army-count, army count overrunning the file, implausible fleet count, truncated fleet/nation table). Their messages can no longer be produced for any input. Harmless, but it is dead code plus a duplicated table walk on every SAV parse. Worth collapsing once the blocking fix settles the shape of `Detect`.

3. **`SaveNationTable.ParseDat` / `SaveRecruitmentTable`'s DAT branch read at fixed offsets with no explicit length guard**, relying on `Detect` having established `data.Length == 140706`. That is sound today but the invariant is implicit. If the blocking fix changes `Detect`'s DAT arm, please add the explicit `tableStart + count * stride <= data.Length` check that `SaveArmyTable.cs:46` and `SaveFleetTable.cs:29` already have, so all four DAT paths guard the same way.

4. **`tests/IC2.Data.Tests/LocalAssets.cs:45` catches bare `Exception`** to decide "not configured". The comment names the three types that mean that, so catch those three rather than everything — as written, a typo'd `assets.local.ini` (or a genuine bug in `AssetSettings.Load`) turns into 65 silent skips on a machine that *is* configured, which is the failure mode DoD 8 is trying to avoid in the other direction. Narrowing the catch costs nothing and keeps the skip honest.

5. Minor: the fixture doc comment should state that expected values were generated by the code under test (see gate 1, DoD 4).

---

### Summary

| Gate | Verdict |
| --- | --- |
| 1. DoD, reproduced by me | **Pass** — all 8 lines run independently; DoD 4 and DoD 6 verified in depth |
| 2. Provenance | **Pass** — every offset traced; derived constant disclosed and independently corroborated |
| 3. Determinism | **Pass** |
| 4. Scope | **Pass** — 18/18 files inside Owns; no plan diff |
| 5. Correctness | **Fail** — one blocking finding in `SaveFormat.Detect` |

One blocking item, with a concrete three-line fix and a repro. Everything else in this PR is in good shape and I expect this to close in a single round.

This is round 1; no disagreement to escalate.

diegoami:
## Review addendum — gate 5, additional correctness findings

The `/code-review --effort high` sweep over `main...HEAD` returned after I posted the review above. It independently reached the same blocking conclusion on `SaveFormat.cs:45` and surfaced **five further findings I had missed, two of them inside T30's own Owns list and both reproducible crashes.** Verdict is unchanged (**request changes**), but the fix list is longer than my first comment implied. I reproduced the two in-scope crashes myself at PR head before relaying them.

### It makes the blocking finding worse, in a way I under-called

I argued the misclassification fails loudly. That is true of the nation path but **not** of the fleet path: `SaveFleetTable.cs:24-31`'s DAT branch does **no field validation at all**, so a 140,706-byte SAV yields **two fabricated `FleetRecord`s read out of the middle of the SAV's army table**, with no error. So this is a genuinely silent-wrong-data path, not merely a confusing-error path. That strengthens the case for the content anchor I suggested (require `NationCatalog.Name(0)` at `DatLayout.NationTableStart` before returning `Dat`), or for trying the SAV walk first and only falling back to length.

### New — in scope, reproducible crash: `src/IC2.Inspect/SaveJsonExporter.cs:11,34`

`Program.cs:52-55` was updated for the new `Leader`/`HumanPlayer` model, but **the second consumer in the same project was not**. Reproduced at PR head:

```
$ dotnet run --project src/IC2.Inspect -- --to-json "Imperial Conquest 2.dat" out.json
Unhandled exception. IC2.Data.DatDataNotPresentException: The DAT has no calendar/current-turn trailer at all …
   at IC2.Data.SaveTurnState.Parse(...) SaveTurnState.cs:line 36
   at SaveJsonExporter.Export(...) SaveJsonExporter.cs:line 11
   at Program.<Main>$(...) Program.cs:line 153
```

and past that, `SaveJsonExporter.cs:34`'s `humanPlayer = n.HumanPlayer` would throw too. The PR body says "`--to-json` still throws unhandled on the DAT (it always did, via a different exception before this change) — exporting the DAT to JSON is T29's job". I accept that **exporting** the DAT is T29's job and should not be built here. But "it always did" is not the right test: this PR changed the exception type to one nothing catches, and the file is inside `src/IC2.Inspect/**`. The in-scope fix is not to implement DAT export — it is to fail *cleanly* with the message instead of a stack trace, which is a one-line guard.

### New — in scope, reproducible crash: `src/IC2.Inspect/Program.cs:12, :174`

`DatDataNotPresentException` derives from `InvalidOperationException`, and `UnrecognizedSaveFormatException` from `Exception`. **Neither is matched by any of the CLI's eleven `catch (Exception ex) when (ex is IOException or UnauthorizedAccessException or ArgumentException)` filters.** Reproduced:

```
$ dotnet run --project src/IC2.Inspect -- --list-mercenaries "Imperial Conquest 2.dat"
Unhandled exception. IC2.Data.DatDataNotPresentException: The DAT has no mercenary-offer pool. …
   at IC2.Data.SaveMercenaryTable.Parse(...) SaveMercenaryTable.cs:line 41
   at Program.<Main>$(...) Program.cs:line 174
```

Same for `--inspect-turn` on the DAT (`Program.cs:12`). The exception messages the implementer wrote are genuinely good — they explain exactly why the value is absent and cite the investigation — and the user never sees them, because they arrive as an unhandled stack trace with a non-zero runtime exit code instead of the clean `stderr` + `return 1` path every other error takes. Adding `DatDataNotPresentException` and `UnrecognizedSaveFormatException` to those filters is squarely inside Owns and makes the absence-modelling work end to end. This is the single highest value-per-line fix in the list.

### New — `src/IC2.Data/SaveArmyTable.cs:76`: the tombstone skip has no upper bound

A file whose owner words **all** read `0xFFFF` now "parses cleanly" to `Armies.Count == 0` with N skipped records and no error, where previously any `0xFFFF` owner aborted. `--list-armies` would print "No armies owned by X" rather than reporting corruption. That is the "tombstone tolerance must not become blanket tolerance for corruption" edge of DoD 3, approached in bulk rather than per-record. The PR's own fixture already asserts `SkippedArmies <= 1` across all 51 saves (`CorpusSweepTests.cs:105`), so the evidence for a sanity cap — or at minimum an error when *every* record is a tombstone — is already in the branch. Cheap, and it closes the gap without widening anything.

### New — `src/IC2.Data/DatLayout.cs:44-45`: the provenance prose is off by one

The comment says the 320-byte recruitment queue at 713 ends at 1,033, "one byte before the unlabelled 4-byte field that precedes treasury at 0x40D (1,037)". It is **zero** bytes before: 1,033 is exactly where that 4-byte field begins (1,033–1,036), and 1,037 = `0x40D` follows it. **The constant itself is correct** — I verified the full chain independently in my gate 2 notes above, and confirmed `0x2C9` empirically against the SAV path — so this is a prose error, not a data error. Flagging it only because this is the one file in the PR whose stated purpose is auditable derivation; an off-by-one in the audit trail is the kind of thing §4.5 exists to keep clean. Fix the sentence, keep the constant.

### New — `tests/IC2.Data.Tests/CorpusFixtures/CorpusSweepTests.cs:89`

`Directory.GetFiles` over three hard-coded subdirectories throws `DirectoryNotFoundException` on a configured machine that happens to lack one of them (say `saves-processed/processed`) — a hard failure rather than the explicit skip DoD 8 asks for. `LocalAssets.IsConfigured` only verifies the DAT is present, so this is reachable. Guard the three directories with `Directory.Exists`.

### New — `tests/IC2.Data.Tests/DatFormatTests.cs:84`: the DoD 6 negative test does not pin the contract

This one matters more than its size suggests, because DoD 6 is precisely the "never a fabricated default" line:

```csharp
Assert.True(string.IsNullOrEmpty(nation.Leader), ...);
```

`IsNullOrEmpty` **also passes for `""`** — the exact fabricated-empty-string outcome the property's own doc comment says it exists to prevent ("Modelled as an explicit absence rather than an empty string, precisely so a caller cannot mistake 'not stored' for 'stored and blank'"). The **implementation is correct** — I read it, `SaveNationTable.cs:63` passes literal `leader: null` — so DoD 6 is genuinely met today, and this is not why I am requesting changes. But the test guarding it would not notice if a future change swapped `null` for `""`. Please make it `Assert.Null(nation.Leader)`.

---

### Revised gate 5 fix list

| # | File | Severity |
| --- | --- | --- |
| 1 | `src/IC2.Data/SaveFormat.cs:45` — length-only DAT detection; silent wrong data via `SaveFleetTable`'s unvalidated DAT branch | **Blocking** |
| 2 | `src/IC2.Inspect/Program.cs:12,174` — CLI catch filters miss both new exception types; stack traces instead of messages | High, in Owns |
| 3 | `src/IC2.Inspect/SaveJsonExporter.cs:11,34` — second consumer not updated; unhandled crash on the DAT | Medium, in Owns |
| 4 | `src/IC2.Data/SaveArmyTable.cs:76` — unbounded tombstone skip; all-tombstone file parses "clean" | Medium |
| 5 | `src/IC2.Data/SaveTurnState.cs:35` + `SaveArmyTable.cs:51-56` + `SaveFleetTable.cs:34-48` — exception-type contract change and now-dead guards (my note 2, corroborated) | Low |
| 6 | `src/IC2.Data/DatLayout.cs:44-45` — off-by-one in the derivation prose (constant is correct) | Low |
| 7 | `tests/IC2.Data.Tests/CorpusFixtures/CorpusSweepTests.cs:89` — hard failure instead of skip on a partial corpus | Low |
| 8 | `tests/IC2.Data.Tests/DatFormatTests.cs:84` — `Assert.Null`, not `IsNullOrEmpty`, to pin DoD 6 | Low |
| 9 | `tests/IC2.Data.Tests/LocalAssets.cs:45` — narrow the bare `catch (Exception)` | Low |

Gates 1–4 still pass exactly as written in my first comment — nothing here touches the DoD reproduction, the provenance verdict, determinism, or scope. Items 2 and 3 are both inside `src/IC2.Inspect/**` and so are squarely in scope to fix on this branch; none of the nine requires reaching outside Owns.

Label set to `status:rework`. This is still round 1 and there is no disagreement between me and the implementer to escalate — every item above is either reproducible at PR head or a one-line test/comment tightening.

diegoami:
Pushed fixes for all three review findings (commit 4dcdc3f):

1. **Blocking** — `SaveFormat.Detect` now tries the SAV count-word structural walk (`SaveNationLayout.Locate`) *first* and only falls back to the DAT length match (140,706 bytes) when that walk fails. A genuine DAT still reliably fails the SAV walk (army 0's X word gets misread as a SAV army count, overrunning the file — the exact cascade `docs/investigations/dat-file-layout.md` already traces), so this is a pure reorder, not a new heuristic. Added `DatFormatTests.A_real_sized_sav_padded_to_the_dats_exact_length_is_still_detected_and_parsed_as_sav` (reproduces your padded-SAV repro directly, via `SyntheticSaveBuilder.PadTo`, no real assets needed) plus a paired `An_all_zero_file_the_exact_dat_length_is_a_degenerate_but_valid_empty_sav` test pinning the other edge case honestly (an all-zero 140,706-byte file is a structurally valid *empty* SAV, not a DAT — I didn't paper over that). `SaveFleetTable`'s DAT branch doesn't need separate validation now that it can only ever be reached for a file that has already failed the SAV walk.
2. `IC2.Inspect/Program.cs`'s eleven catch filters now also match `DatDataNotPresentException` and `UnrecognizedSaveFormatException`. Verified `--list-mercenaries` and `--inspect-turn` against the real DAT both now print the actual message (exit 1), not a stack trace.
3. `SaveJsonExporter` (`--to-json`) no longer crashes on the DAT: `SaveTurnState`/`SaveMercenaryTable` are now read defensively (`turn: null`, `mercenaryOffers: []` on the DAT), and `humanPlayer` is `null` rather than touching the throwing accessor. Verified end-to-end against the real DAT (`turn: None`, `mercenaryOffers: []`, Rome's nation data correct) and confirmed a normal SAV export is byte-for-byte unchanged (real leader name, real `humanPlayer`, real turn state).

Full suite: `dotnet build IC2.sln` clean, `dotnet test IC2.sln` 234/234 passed (78 IC2.Data.Tests + 156 IC2.Engine.Tests), and the assets-absent skip path re-verified (13 pure-synthetic pass, 65 asset-dependent skip explicitly).

diegoami:
## Review round 2 — T30 (#37), PR #39 @ `4dcdc3f` · Opus / Medium

**Verdict: approve.** All five gates pass. The blocking finding and both in-scope crashes are genuinely fixed — I re-ran my own round-1 reproductions rather than reading the PR comment, and verified the `--to-json` claim by byte-diffing against a build of `main`.

Per §6.5 I am **not** smoothing anything over, so read the "Still open" section: **six of my nine round-1 findings are untouched, and this round introduced one new (minor) issue.** I do not consider any of them merge-blocking and I explain why below — but they are real, and it looks like only three of the nine were relayed to the implementer, so this is an orchestration gap rather than the implementer declining to act.

---

### Finding 1 (blocking) — `SaveFormat.Detect` — **fixed, verified**

The fix reverses the order: `SaveNationLayout.Locate` is tried first and is authoritative when it succeeds; the 140,706-byte length match is now a fallback in the `catch`. That is the right shape — a structural check beats a coincidence-prone length check.

I re-ran my round-1 reproduction **unchanged** (same real save padded to exactly `DatFileLength`):

```
padded file length: 140706  (DAT length = 140706)

--- padded SAV, round-2 head ---
Rome · computer player · leader Lucius Papirius · capital Rome
25 cities (map count 25) · … tax 10% · mobilized 30% · treasury 1200 talents · unity value 821
Army 0 at (97, 33) · 23,700 troops · 237 tons supply (100%) · 600 money · moves 0 · morale 70
Army 1 at (115, 49) · 22,000 troops · 220 tons supply (100%) · 600 money · moves 0 · morale 68

--- the SAME file unpadded, for comparison ---
Rome · computer player · leader Lucius Papirius · capital Rome
25 cities (map count 25) · … tax 10% · mobilized 30% · treasury 1200 talents · unity value 821
Army 0 at (97, 33) · 23,700 troops · 237 tons supply (100%) · 600 money · moves 0 · morale 70
Army 1 at (115, 49) · 22,000 troops · 220 tons supply (100%) · 600 money · moves 0 · morale 68
```

Byte-identical. Round 1 produced `Army 0 has out-of-range owner code 33` here. Fixed, and the regression test `A_real_sized_sav_padded_to_the_dats_exact_length_is_still_detected_and_parsed_as_sav` reproduces it without needing local assets, so CI keeps it honest.

**The other direction also holds**, which is the part the reorder could have broken. A genuine DAT must still *fail* the SAV walk or it would now be misclassified as a SAV. It does, and for a documented reason rather than luck: `dat-file-layout.md:169-171` already traces exactly this cascade — army 0's X word (100) is misread as a SAV army count, driving the computed fleet-count offset to 166,558, past the end of the 140,706-byte file, so `Locate` throws *"Save ends before the fleet count"*. I confirmed it end to end against the real DAT: `--inspect-nation` still reports `not stored in the DAT` with the DAT's own treasury 2200 (vs 1200 in the save, so it is genuinely reading DAT offsets), and `--list-fleets` returns Carthage 90 ships / 80 tons / condition 85 — the exact values `dat-file-layout.md:82-84` cites as its cross-check. The comment in `SaveFormat.cs` explains this correctly and cites the doc.

**On the 0xFF-fill question — the reasoning holds, I checked it rather than accepting it.** Under the new ordering an all-zero 140,706-byte array reads as `armyCount = 0`, `fleetCountOffset = 100,958` (in range), `fleetCount = 0`, nation table at 100,960 needing 18,752 bytes, ending at 119,712 — comfortably inside the file. So `Locate` **succeeds**, and an all-zero file of that length genuinely *is* a structurally valid (degenerate, empty) SAV. The old all-zero fixture would therefore have started failing for a correct reason, and switching it to 0xFF (`armyCount = 65535` → computed offset ~43 MB → walk fails → falls back to length → `Dat`) is a legitimate fixture repair, not a test bent to fit the code. The implementer also added `An_all_zero_file_the_exact_dat_length_is_a_degenerate_but_valid_empty_sav` pinning that edge explicitly instead of quietly dropping it. That is the right way to handle it, and I'd have objected if the all-zero case had simply disappeared.

### Finding 2 — `Program.cs` catch filters — **fixed, verified**

I audited all of them mechanically rather than counting by eye: **12** `catch (Exception ex) when (…)` filters in the file (I said 11 in round 1 — there are 12, including the trailing top-level one), and grep confirms **zero** now lack either `DatDataNotPresentException` or `UnrecognizedSaveFormatException`. Run against the real DAT:

```
$ --list-mercenaries "Imperial Conquest 2.dat"
The DAT has no mercenary-offer pool. It is New Game state the original has not created yet … 
exit=1

$ --inspect-turn "Imperial Conquest 2.dat"
The DAT has no calendar/current-turn trailer at all — no code assigns starting week/season/year/… 
exit=1
```

Clean message on stderr, exit 1, no stack trace, no unhandled-exception exit code. The carefully written absence messages now actually reach the user, which was the point.

### Finding 3 — `SaveJsonExporter` — **fixed, and the SAV path verified unchanged**

`--to-json` on the DAT now exits 0 and writes a document. More importantly, I did not take "byte-for-byte unchanged" on trust — I created a `git worktree` at `main`, built it, exported the same save with both builds, and diffed:

```
233554  t30_sav_base.json   (main)
233554  t30_sav_new.json    (4dcdc3f)
$ diff t30_sav_base.json t30_sav_new.json
IDENTICAL - no change to SAV export
```

So the anonymous-type restructuring (`turn` moving behind an `object?`) did not perturb property order, nullability, or values for a normal save. On the DAT, `turn` serialises as `null` and `leader`/`humanPlayer` as `null` — honest absences rather than fabricated values, consistent with how the parsers model them.

---

### Gates

| Gate | Verdict |
| --- | --- |
| 1. DoD, reproduced by me at `4dcdc3f` | **Pass** |
| 2. Provenance | **Pass** |
| 3. Determinism | **Pass** |
| 4. Scope | **Pass** |
| 5. Correctness | **Pass** (with open items, below) |

**Gate 1** re-run at *this* commit, not carried over: `--list-armies … Thracia` still gives `98 tons (44%) · morale 66` and `0 tons (0%) · morale 51`, matching `thracia-supply-morale.md`. Full suite `dotnet build IC2.sln` clean 0 warnings, `IC2.Data.Tests 78/78`, `IC2.Engine.Tests 156/156`. DoD 8 re-verified by renaming `assets.local.ini`: `Passed: 13, Skipped: 65, Total: 78` (13 rather than round 1's 11 because the two new discriminator tests are pure-synthetic — correct). Config restored; working tree clean.

**Gate 2**: the one new factual claim this round — that a genuine DAT reliably fails the SAV walk — traces to `dat-file-layout.md:169-171` and I verified it against the real file. No new constants introduced. **Gate 3**: no randomness, clock, or ordering added. **Gate 4**: all 5 changed files inside `src/IC2.Data/**`, `src/IC2.Inspect/**`, `tests/IC2.Data.Tests/**`; no diff to `docs/build-orchestration-plan.md`.

---

### Still open — stated plainly, not smoothed over

Round 1 filed nine items. Three were fixed. **Six were not touched**, and from the round-1 response it looks like only those three were relayed — so this reads as an orchestration gap, not the implementer ignoring review. Verified still present at `4dcdc3f`:

| # | Item | Severity |
| --- | --- | --- |
| 4 | `SaveArmyTable.cs:76` — unbounded tombstone skip; an all-`0xFFFF` file parses "clean" to `Armies.Count == 0` with no error | Medium |
| 5 | `SaveArmyTable.cs:52,56` / `SaveFleetTable.cs:34-48` — SAV-branch guards still unreachable (`Detect` returns `Sav` only when `Locate` already succeeded), plus the `InvalidDataException` → `UnrecognizedSaveFormatException` contract change that breaks `godot/MapViewer.cs:389-393`'s graceful degradation (out of Owns — needs routing to the Godot lane) | Low |
| 6 | `DatLayout.cs:38` — "one byte before" is off by one; should be zero bytes (1,033 is where that 4-byte field *begins*). Constant is correct, prose is not | Low |
| 7 | `CorpusSweepTests.cs:89` — `Directory.GetFiles` over hard-coded subdirs hard-fails instead of skipping on a partial corpus | Low |
| 8 | `DatFormatTests.cs:84` — `IsNullOrEmpty` also passes for `""`, so the DoD 6 negative test does not pin the null contract the implementation actually honours; `Assert.Null` would | Low |
| 9 | `LocalAssets.cs:45` — bare `catch (Exception)` can turn a real misconfiguration into 65 silent skips | Low |

**New this round:** `SaveJsonExporter.cs` emits `mercenaryOffers: []` for the DAT. I confirmed it — the DAT export carries an empty array, not null. That is indistinguishable from a real save whose offers are all empty, and it is precisely the *"fabricated 'empty pool'"* that the `DatDataNotPresentException` message being caught two lines earlier says must not be invented. `turn`, `leader` and `humanPlayer` in the same function all correctly use `null`; this one field breaks the pattern. Two-line fix: make the local nullable and emit `null`. Severity: low — `--to-json` is a debug dump, T29 writes its own exporter against T02's schema and does not consume this shape.

**Why none of this blocks the merge.** Item 4 guards against corruption that does not exist in the 51-save corpus (the fixture asserts ≤ 1 tombstone per file); items 5–9 are the ones I labelled non-blocking in round 1 and it would be moving the goalposts to gate on them now, especially having apparently never asked; and the new `mercenaryOffers` issue touches only a debug tool that nothing downstream reads. **None of them affects what T30 exists to unblock**: the two defect classes are correctly handled, the full 52-file corpus sweep is green, and T29's DAT parse and T21's save imports are both satisfied. Spending the last rework round before human escalation on a prose typo and five Lows would be the wrong trade.

**Recommendation for the orchestrator:** merge this, and either relay items 4–9 plus the `mercenaryOffers` nit to the implementer as a quick pre-merge touch-up (all are small and all are inside Owns except item 5's `MapViewer.cs` half), or file them as a follow-up issue against `IC2.Data`/`IC2.Inspect`. Item 5's `MapViewer.cs` half needs the Godot lane regardless, since it is outside every T30 Owns entry.

Label set to `status:approved`. No disagreement between me and the implementer at any point — round 1's findings were accepted and fixed as filed, and I verified each one independently rather than by reading the response. Nothing here needs human escalation.

Issue #3, opened 2026-09-13
T03 Engine seams


diegoami:
**Design milestone:** none explicitly (implied by `game-design.md` §Testing and determinism) · **Branch:** `task/T03-engine-seams`
**Model / effort:** **Opus / Ultrahigh** · **Reviewer:** Opus / High **+ `/code-review --effort ultra`**
**Owns:** `src/IC2.Engine/Core/**`, `tests/IC2.Engine.Tests/Core/**`

### Scope
The interfaces every later task is written against, and nothing else:
- `IRng` — one seeded service threaded through the turn coordinator; the only source of randomness in gameplay code.
- The turn coordinator: an ordered phase pipeline with a stable declared phase order and an `OnQuarterBoundary` hook **contract** (fired later by T06's calendar, subscribed by T08 and T19), so economy tests can fire it without a calendar merged.
- Command dispatch: `ICommand` → `CommandResult` with a typed rejection. An illegal command never throws and never silently no-ops.
- The domain-event sink that T10's news log and the UI both consume.
- **Attribute-based system registration** (assembly scan) so adding a system touches only that system's own file. This is the plan's main anti-conflict mechanism.
- The determinism guard test.

### Done when
- [ ] Two runs of a scripted N-phase sequence with the same seed produce byte-identical `GameState` hashes; different seeds produce different hashes
- [ ] A no-op system registers by attribute alone and runs in declared phase order; a second one added in a second file requires editing no shared file
- [ ] An illegal command returns a typed rejection with a reason code — no exception thrown, no state changed
- [ ] The determinism guard test fails on a deliberately-added `new Random()` under `src/IC2.Engine` (prove the guard works, then remove it)
- [ ] `OnQuarterBoundary` can be fired directly in a test with no calendar implementation present

### Blocked by
- [ ] #2 T02 Core domain model

### Hazard
**The critical-path blocker.** No other task may be in flight while this is open — its diff defines the interfaces everything else will compile against.

---
Full task entry (the binding contract): [`docs/build-orchestration-plan.md` → T03 Engine seams](https://github.com/diegoami/imperial_conquest_2/blob/main/docs/build-orchestration-plan.md#t03-engine-seams-rng-turn-pipeline-commands-events)

An implementer may **not** weaken a "Done when" line, skip a test, or edit the plan document — it escalates instead (plan §6.4). Review contract: plan §6.2.

PR #44, opened 2026-09-13
T07: Strength functions

### LINKED ISSUE #7: T07 Strength functions
**Design milestone:** M5 · **Branch:** `task/T07-strength-functions`
**Model / effort:** Sonnet / High · **Reviewer:** **Opus / Medium**
**Owns:** `src/IC2.Engine/Strength/**`, `tests/IC2.Engine.Tests/Strength/**`

### Scope
The three pure functions that M8, M9 and M12 all consume, extracted early exactly as `game-design.md` M5 says: `armyPower`, `fleetPower`, and siege defender strength. **Integer semantics are part of the specification** — the original is Delphi and truncates; every division must be pinned by a test rather than left to C# operator defaults.

### Done when
- [ ] `armyPower(a) = (Σ powerWeight[type] × troops / 100) / 80 × armyMorale` reproduces hand-computed values for the published 13-unit Roman roster, using the `+0x26` weights from the fixtures corpus
- [ ] `fleetPower(f) = ships × condition / 10 (+ carriedArmyPower / 50)` reproduces hand-computed values, both with and without the carried-army term
- [ ] The `× (1 + random(4)/10)` bonus goes through `IRng` and is exactly reproducible under a fixed seed
- [ ] Siege defender strength triples archers and applies the fortification/loyalty term
- [ ] At least three truncation cases where integer division differs from floating-point division assert the integer result

### Blocked by
- [ ] #3 T03 Engine seams
- [ ] #4 T04 Fixtures corpus

### Hazard
`design-audit.md` §2.9 — **there are two different morales**. Only army record `+14` (strategic) feeds these functions; the per-unit tactical morale array must not appear in this diff.

---
Full task entry (the binding contract): [`docs/build-orchestration-plan.md` → T07 Strength functions](https://github.com/diegoami/imperial_conquest_2/blob/main/docs/build-orchestration-plan.md#t07-strength-functions)

diegoami:
Closes #7

## Task

- **Task**: T07 Strength functions — [plan entry](https://github.com/diegoami/imperial_conquest_2/blob/main/docs/build-orchestration-plan.md#t07-strength-functions)
- **Branch**: `task/T07-strength-functions`

| --- | --- |
| `src/IC2.Engine/Strength/ArmyPower.cs` | `src/IC2.Engine/Strength/**` |
| `src/IC2.Engine/Strength/FleetPower.cs` | `src/IC2.Engine/Strength/**` |
| `src/IC2.Engine/Strength/SiegeStrength.cs` | `src/IC2.Engine/Strength/**` |
| `tests/IC2.Engine.Tests/Strength/ArmyPowerTests.cs` | `tests/IC2.Engine.Tests/Strength/**` |
| `tests/IC2.Engine.Tests/Strength/FixedDrawRng.cs` | `tests/IC2.Engine.Tests/Strength/**` |
| `tests/IC2.Engine.Tests/Strength/FleetPowerTests.cs` | `tests/IC2.Engine.Tests/Strength/**` |
| `tests/IC2.Engine.Tests/Strength/RulesetMatchesFixtureCorpusTests.cs` | `tests/IC2.Engine.Tests/Strength/**` |
| `tests/IC2.Engine.Tests/Strength/SiegeStrengthTests.cs` | `tests/IC2.Engine.Tests/Strength/**` |
| `tests/IC2.Engine.Tests/Strength/StrengthTestbed.cs` | `tests/IC2.Engine.Tests/Strength/**` |

- [x] Every changed file above is inside the task's declared **Owns** list. `Ruleset.Siege` itself (the round-1 blocking finding) was corrected by **T31** (`27d7b41`, merged to `main`), not by this PR — this branch was rebased onto it, per plan §6.7.
- [x] No change to `docs/build-orchestration-plan.md`, original game files, `assets.local.ini`, `.gitignore`'s exclusion policy, or anything under `docs/reports/`.

## DoD evidence

```text
DoD 1: armyPower(a) = (Σ powerWeight[type] × troops / 100) / 80 × armyMorale reproduces
       hand-computed values for the published 13-unit Roman roster, using the +0x26 weights
       from the T04 corpus.
$ dotnet test tests/IC2.Engine.Tests/IC2.Engine.Tests.csproj \
    --filter "FullyQualifiedName~ArmyPowerTests.Compute_Roman13UnitRoster_MatchesHandComputedValue"
Total tests: 1
     Passed: 1

DoD 2: fleetPower(f) = ships × condition / 10 (+ carriedArmyPower / 50) reproduces
       hand-computed values, including the carried-army term and its absence.
$ dotnet test tests/IC2.Engine.Tests/IC2.Engine.Tests.csproj \
    --filter "FullyQualifiedName~FleetPowerTests.Compute_WithoutCarriedArmy_MatchesHandComputedBase|FullyQualifiedName~FleetPowerTests.Compute_WithCarriedArmy_MatchesHandComputedBase"
Total tests: 2
     Passed: 2

DoD 3: The × (1 + random(4)/10) bonus is applied through IRng and is exactly reproducible
       under a fixed seed (same seed → same value, asserted twice in one test).
$ dotnet test tests/IC2.Engine.Tests/IC2.Engine.Tests.csproj \
    --filter "FullyQualifiedName~Compute_RandomBonus_IsReproducibleUnderTheSameSeed"
Total tests: 1
     Passed: 1

DoD 4: Siege defender strength triples archers, and applies the fortification/loyalty term.
$ dotnet test tests/IC2.Engine.Tests/IC2.Engine.Tests.csproj \
    --filter "FullyQualifiedName~SiegeStrengthTests"
Total tests: 8
     Passed: 8

DoD 5: A truncation test: at least three cases where integer division differs from
       floating-point division assert the integer result.
$ dotnet test tests/IC2.Engine.Tests/IC2.Engine.Tests.csproj \
    --filter "FullyQualifiedName~PerUnitTruncation|FullyQualifiedName~FinalDivisionTruncation|FullyQualifiedName~RandomBonusTruncationOrder"
Total tests: 3
     Passed: 3

DoD 6: armyPower is exercised across the full confirmed 51…70 morale range and at both
       bounds, and the strength it returns is monotonic in morale over that range.
$ dotnet test tests/IC2.Engine.Tests/IC2.Engine.Tests.csproj \
    --filter "FullyQualifiedName~AcrossConfirmedMoraleRange"
Total tests: 1
     Passed: 1

Whole solution:
$ dotnet build IC2.sln
Build succeeded. 0 Warning(s). 0 Error(s).

$ dotnet test IC2.sln
IC2.Engine.Tests: Passed! - Failed: 0, Passed: 193, Skipped: 0, Total: 193
IC2.Data.Tests:   Failed: 3, Passed: 75, Skipped: 0, Total: 78
  (all 3 failures are pre-existing, local-asset-dependent (T30's saves/ corpus drift on this
   machine), entirely outside this PR's Owns list -- see "Notes for the reviewer". `git diff
   --stat main...HEAD` confirms no changes outside src/IC2.Engine/Strength/** and
   tests/IC2.Engine.Tests/Strength/**.)
```

## Provenance checklist

- [x] Every constant introduced or changed traces to a `tests/fixtures` entry or a cited report under `docs/reports/`.
  - `+0x26` power weights: read from `Ruleset.UnitTypes[*].CombatPowerWeight` (T02's shipped `toy-ruleset.json`), cross-checked in `RulesetMatchesFixtureCorpusTests` against the T04 corpus's `unitType.*.powerWeight` fixtures (all `decompiled-unit-map-orders-and-record-fields.md`).
  - `CombatRules.PowerTroopDivisor` (100), `PowerDivisor` (80), `NavalCombatRules.ConditionDivisor` (10), `CarriedArmyPowerDivisor` (50), `RandomBandCount` (4), `RandomBandPercent` (10), `SiegeRules.ArcherStrengthMultiplier` (3), `Siege.PowerDivisor` (80) — all in T02's shipped `Ruleset` model, provenance-tagged there; this PR reads them, invents none.
  - `SiegeRules.DefenderLoyaltyWeight` (150), `DefenderFortificationWeight` (250), `DefenderPopulationWeight` (200) — **corrected by T31** (`docs/investigations/siege-defender-strength.md`) after round-1 review found the original names/weights swapped and the third field misnamed "unidentified". This PR now reads the corrected names and decodes the fortification word through `FortificationCode.FinishedPercent`, per T31's own `SiegeRules` doc comments.
  - The 13-unit Roman roster (troops/types): `army-records-and-roman-roster.md`, matching the T04 corpus's `roman13.troops` fixture (48,173).
  - `ArmyManagementRules.NewArmyMorale` (59), used as the test's morale input rather than an arbitrary literal.
- [x] Every `[designed]` value states what was searched and came up empty before it was designed — **n/a, no `[designed]` value in this diff**; every number is `[confirmed]`.

Values omitted because no source was found (or "none"):

- The siege-defender formula's `×5/3` high-fortification bonus and its capital-only gate (`FUN_0044B8D0`, now identified but its ruleset fields are misnamed — filed as [issue #46](https://github.com/diegoami/imperial_conquest_2/issues/46)), and the owner/allegiance `×4/5` penalty (filed as [issue #47](https://github.com/diegoami/imperial_conquest_2/issues/47), a shape mismatch between the ruleset field and the actual operation). Both are confirmed in `FUN_0044A98C` but are open bugs against `Ruleset.Siege`, outside this task's Owns list and outside DoD 4's literal text — not worked around here.
- The garrison-troops addend (`SiegeRules.DefenderGarrisonTroopDivisor`). **Correction from round 1**: this is *not* assigned to T17 in any provenance note (unlike the two items above) — it is omitted here because computing it needs per-nation recruitment-slot state this pure function does not take as an input, not because any task owns resolving an open identity.

## Determinism

- [x] No `System.Random`, wall-clock reads, `Guid.NewGuid`, or order-dependent iteration in gameplay paths; every random draw goes through `IRng`. `FleetPower.Compute` takes `IRng` and calls `NextInt(rules.RandomBandCount)`; `FleetPowerTests.Compute_RandomBonus_IsReproducibleUnderTheSameSeed` proves same-seed reproducibility with `SplitMix64Rng` (the engine's real implementation) by replaying the same seed's own draw independently and hand-verifying the result, rather than relying on a second seed happening to differ (round-1 finding — see below). `FixedDrawRng` (a test-only `IRng` stub used to isolate arithmetic in the other FleetPower tests) lives in the test project, not `src/`, and `ArmyPower`/`SiegeStrength` draw no randomness at all.

### Round 1 rework — every finding addressed

**Rebase.** This branch is rebased onto `main` at `27d7b41` (T31, which corrected `Ruleset.Siege`'s defender-strength field identities per the round-1 review's blocking finding). No change to `Ruleset.Siege` is made by this PR.

**Blocking finding — `SiegeStrength.Defender`'s field/weight mapping.** Rewritten against T31's corrected names: `loyalty × DefenderLoyaltyWeight (150) + finishedFortificationPercent × DefenderFortificationWeight (250) + populationThousands × DefenderPopulationWeight (200)`. The fortification word is now decoded through `FortificationCode.FinishedPercent` (the guarded `<= 100 ? code : code % radix` rule) before weighting, never raw and never an unguarded `% 100`. No parameter or doc comment calls anything "unidentified" or tells a caller to pass 0. New tests: `Defender_AppliesLoyaltyFortificationAndPopulationTerms`, `Defender_DecodesInProgressFortificationWord` (a code of 130 must weight as 30, not 130), `Defender_FinishedFortificationAtExactlyMaximum_IsNotZeroed` (a stored 100 must weight as 100, not 0 — the specific failure mode an unguarded `% 100` would cause), `Defender_FormulaShape_MatchesCorpusFixtureWording`.

**Gate-5 correctness bug — `FleetPower.cs`'s `RandomBandPercent` used as a divisor.** Fixed: the bonus is now `baseValue * rules.RandomBandPercent / 100` (computed as one integer division), which reduces to exactly the same `floor(baseValue / 10)` for the shipped value of 10 — so no existing DoD result changes — but now scales correctly if the ruleset value is ever changed. New regression test `Compute_RandomBandPercent_IsAPercentageNotADivisor` constructs a ruleset with `RandomBandPercent = 20` and asserts the old (buggy) reading is *not* what comes out.

**The 8 minor findings, each addressed:**
1. Dead assertion in `Attacker_NoDivisionByHundred_UnlikeArmyPower` — deleted; kept only the case that can actually fail for the stated reason.
2. DoD-5 case 1 was missing the float-parity assertion its siblings have — added (`Assert.NotEqual((int)Math.Round(...), power)`).
3. `Compute_RandomBonus_IsReproducibleUnderTheSameSeed`'s different-seed check was a 1-in-4 flake risk — replaced with a direct hand-verification against the same seed's own replayed draw (no second seed needed at all).
4. `SiegeStrength.Attacker` never validated `archerUnitTypeId` — now throws `ArgumentException` on an unrecognised id, matching `ArmyPower.Compute`'s existing behaviour. New test `Attacker_UnknownArcherUnitTypeId_Throws`.
5. `FleetPower.CarriedArmyStrength` was a `record struct`, whose `default` is a non-null value that would reach `SiegeStrength.Attacker` with a null `Units` — changed to a `sealed record` (reference type), so `null` is the only way to mean "no carried army".
6. `FixedDrawRng.NextInt` ignored `exclusiveUpperBound` — now throws if the fixed draw wouldn't fit, so a stub built for one `RandomBandCount` can't silently assert a value the real `IRng` contract forbids for another.
7. The old `SiegeStrengthTests` comment claimed a match against the T04 corpus fixture that nothing actually checked (a prose string) — `Defender_FormulaShape_MatchesCorpusFixtureWording` now makes a real `Assert.Contains` check against the fixture text.
8. The garrison-troops addend's justification was overstated (see "Provenance checklist" above) — corrected to say it needs state outside this function's inputs, not that T17 owns resolving it.

No morale-mutation logic is implemented anywhere in this diff, and the per-unit tactical morale array (`DAT_004A0350`) is not referenced — only `ArmyState.Morale` (army record `+14`) is consumed, as plain input, never written or clamped.


diegoami:
## T07 review — reviewer: Opus / Medium

Reviewed at PR head (`4c83d99`) checked out in the main checkout. `dotnet build IC2.sln` → 0 warnings / 0 errors; `dotnet test tests/IC2.Engine.Tests` → **185 passed, 0 failed**. CI green, `MERGEABLE`.

**Verdict: request changes (`status:rework`), plus one item the orchestrator should escalate rather than hand straight back.**

Four of five gates pass, and two of the three judgement calls the PR body flags are *correct and well-made*. The blocking finding is in gate 2, and it is the one place the PR's own claim of direct decompilation reading does not survive re-derivation.

---

### The two claims the PR asked to be checked — both verified independently

I re-derived both against `%LOCALAPPDATA%\ReTools\all_app_functions.txt` rather than taking the PR body's word.

**Claim 1 — `fleetPower`'s carried-army term calls the siege formula, not plain `armyPower`: CONFIRMED. This is a real and valuable correction.**

`FUN_0044aa54` @ dump line 49296–49313:

```c
iVar2 = ((short)(&DAT_0049c27e)[i*0xd] * (short)(&DAT_0049c280)[i*0xd]) / 10;   // ships * condition / 10
if (-1 < (short)(&DAT_0049c282)[i*0xd]) {
  iVar1 = FUN_0044a930((&DAT_0049c282)[i*0xd]);                                  // <-- SIEGE strength, not FUN_0044a8cc
  iVar2 = iVar2 + iVar1 / 0x32;                                                  // / 50
}
iVar1 = FUN_0040284c(4);
return iVar2 + iVar1 * (iVar2 / 10);                                             // NOT base * (1 + draw/10)
```

`FUN_0044a930` is unambiguously the call, and `FUN_0044a8cc` (plain `armyPower`) does not appear anywhere in `FUN_0044aa54`. `FleetPower.cs:84` is right, and the class remarks cite it accurately.

Worth recording beyond this PR: this also contradicts **T02's own shipped provenance**, which says of `combat.naval.carriedArmyPowerDivisor` — *"confirmed: decompiled-diplomacy-peace-terms-and-instant-battles.md — a carried army adds armyPower / 50."* It adds *siege* strength / 50. The ruleset note should be corrected when someone owns that file; flagging it here so it isn't lost.

The second half of the same finding — `base + draw * (base / 10)`, with `base / 10` truncated **before** the multiply — is equally confirmed on line 49312, and `FleetPowerTests.cs:89` pins it properly (integer 20 vs. floating-point 22). Good catch; reading this as `base * (1 + draw/10)` in C# integer arithmetic would have silently deleted the entire bonus.

**Claim 2 — DoD 4 describes two distinct decompiled functions, not one: CONFIRMED, and the split does not weaken the DoD.**

Three independent confirmations:
- The dump: `FUN_0044a930` @ 49220 (archer tripling, `type == 2`, `(total / 0x50) * morale`) and `FUN_0044a98c` @ 49249 (fortification/loyalty/third weighted sum) are separate functions with different parameters — an army index and a city index.
- `decompiled-city-capture-resolution.md:8-9` pseudocode for `FUN_0044b27c`, exactly as quoted.
- T02's merged `Ruleset.Siege` genuinely does mirror the split (bare `ArcherStrengthMultiplier`/`PowerDivisor` vs. `Defender*`-prefixed fields) — I checked `data/rulesets/toy-ruleset.json`, the corroboration is real, not post-hoc.

Both clauses of DoD 4 end up implemented and exercised (`SiegeStrengthTests.cs:23` and `:71`), so this is a reconciliation that *strengthens* coverage rather than a reinterpretation that dodges it. Flagging it in the PR body rather than applying it silently was the right call. **No change requested here.**

I also confirmed `FUN_0044a930`'s shape matches `SiegeStrength.Attacker` exactly — no intermediate `/ 100`, non-archers weight 1, `(total / 0x50) * morale` with morale read from `(&DAT_0047c1fa)[i * 0x148]`, i.e. army record **+14**, the strategic field. Correct.

---

### Gate 1 — DoD, independently reproduced: 5 of 6 pass

I ran every line myself rather than trusting the PR body.

| # | Result |
|---|---|
| 1 | **PASS** — I re-did the hand computation independently: per-unit truncated contributions sum to S = 40,506; `40506 / 80 = 506`; `506 × 59 = 29,854`. Roster troops sum to 48,173, matching the corpus `roman13.troops`. Weights used (20/100/60/120) match `+0x26`. |
| 2 | **PASS** — base `50×80/10 = 400`; carried term `((1000×3)/80)×60 / 50 = (37×60)/50 = 2220/50 = 44`; total 444. Both presence and absence covered (`FleetPowerTests.cs:19`, `:42`). |
| 3 | **PASS** — `SplitMix64Rng(12345)` twice → equal; different seed → different. Real `IRng` implementation used, not the stub. |
| 4 | **BLOCKED** — see gate 2. The archer-tripling clause passes cleanly; the fortification/loyalty clause applies the two weights to the wrong fields. |
| 5 | **PASS** — three cases, all green. Cases 2 and 3 are textbook (`ArmyPowerTests.cs:109` asserts 51 vs. float 54; `FleetPowerTests.cs:89` asserts 20 vs. float 22). |
| 6 | **PASS** — `ArmyPowerTests.cs:127` loops the full 51…70 inclusive (20 values), asserts *strict* monotonicity pairwise, and asserts both bounds explicitly (`506×51`, `506×70`) rather than relying on the loop. No clamping added, correctly — that is T08's. |

---

### Gate 2 — Provenance: **FAIL (blocking)**

Most of this gate is genuinely well done. The `+0x26` weights are not hardcoded anywhere: production reads `UnitTypeRules.CombatPowerWeight`, and `RulesetMatchesFixtureCorpusTests.cs:22` cross-checks all five against the T04 corpus's `unitType.*.powerWeight` fixtures. I verified those fixtures against `unit-type-stat-table-in-dat.md`'s table (`+0x26`: 20/100/40/60/120) and `decompiled-unit-map-orders-and-record-fields.md:36`. There are no `[designed]` values in the diff and nothing was invented. That part is exactly right.

**The blocking issue is `SiegeStrength.Defender` (`SiegeStrength.cs:93-102`).**

The PR body states these were *"confirmed in the decompiled code (`FUN_0044A98C`, read directly from the local ReTools dump)"*. I read that function directly, and it does not say what the implementation says.

`FUN_0044a98c`, dump lines 49261-49270:

```c
sVar6 = (&DAT_004795aa)[i*0x11];
if (sVar6 < 0x65) { iVar4 = sVar6; } else { iVar4 = sVar6 % 100; }   // <-- in-progress decode, UNCONDITIONAL
iVar4 = (short)(&DAT_004795a6)[i*0x11] * 0x96     //  * 150
      + iVar4                          * 0xfa     //  * 250  <-- this is the DECODED DAT_004795aa
      + (short)(&DAT_004795ac)[i*0x11] * 200;     //  * 200
if (FUN_0044b8d0(city) && 0x3b < (short)(&DAT_004795a6)[i*0x11]) iVar4 = (iVar4 * 5) / 3;
```

Now the field identities, from `TInformation_ShowCityDetails` @ `0x0043BE5C` (dump lines ~40915-40945) — the *same* panel function `decompiled-unit-map-orders-and-record-fields.md:54` used to name the city record:

- `"Loyalty -"` prints `names[DAT_004795a6 / 10]` → **`DAT_004795a6` is LOYALTY** (the `× 150` field)
- `"Fortification -"` prints `DAT_004795aa`, with `% 100` and `"  (under construction)"` appended when `≥ 0x65` → **`DAT_004795aa` is FORTIFICATION** (the `× 250` field)
- `"Population -"` prints `DAT_004795ac × 1000`, plus `DAT_004795ac × 100 / DAT_004795ae` as *"% of maximum"* → **`DAT_004795ac` is POPULATION IN THOUSANDS** (the `× 200` field)

Independently corroborated by this repo's own merged T02 code: `FortificationCode.AfterSiegeAttempt` (`src/IC2.Engine/Model/FortificationCode.cs:63`) documents `FUN_0044B27C: if (fort > 100) fort = fort % 100`, and `FUN_0044b27c` at dump line 49790 performs exactly that wipe on **`DAT_004795aa`**. So the repo already agrees the fortification word is `DAT_004795aa` — the `× 250` field.

**The actual formula is therefore:**

```
strength = loyalty × 150 + finishedFortificationPercent × 250 + populationThousands × 200
```

**What the PR computes** (`SiegeStrength.cs:99-101`):

```
strength = fortification × 150 + loyalty × 250 + unidentifiedFieldValue × 200
```

Three concrete defects follow:

1. **`fortification` and `loyalty` are swapped against their weights.** A caller passing `city.FortificationCode` to the `fortification` parameter gets it weighted 150 instead of 250, and loyalty weighted 250 instead of 150. This is the first place the mistake becomes an *executable API with parameter names* — T02 only had loosely-named constants, so this PR is where it starts producing wrong numbers for T16/T17.
2. **The third term is population in thousands, not an unidentified field.** `SiegeStrength.cs:87-90` tells callers to *"Pass 0 until T17 resolves what this field is"*, but passing 0 is not neutral — it silently deletes a real term that, for a typical city, dominates the sum. The identity was recoverable from the same panel function in under a minute.
3. **The unconditional `% 100` in-progress decode is missing.** `FUN_0044a98c` always decodes the fortification word before weighting it. Any city with a pending fortification order (stored word > 100) will compute a wildly inflated defender strength. This is not gated on anything T17 owns — `FortificationCode.FinishedPercent` already exists in merged T02 code and does exactly this.

   Related: the doc comment at `SiegeStrength.cs:83-85` attaches the dual in-progress encoding to the **`fortification`** parameter. Given defect 1, that is the parameter carrying loyalty's weight — so a caller who follows the doc decodes the wrong value *and* leaves the real fortification word raw.

Also worth recording for T17: the `× 5/3` branch's threshold tests `DAT_004795a6 > 59`, i.e. **loyalty > 59, not fortification > 59**. T02's `siege.highFortificationThreshold` is mislabelled for the same reason.

**The project already knew this report line was soft.** `docs/design-audit.md` §2.13 lists exactly this as *"One item to re-check rather than trust"* and states plainly: **"`FUN_0044A98C` (defender strength) was not decompiled this pass. [open]"** So the `fortification × 150 + loyalty × 250` line was never a decompilation of that function in the first place — which is precisely why the PR body's claim to have *"read directly from the local ReTools dump"* is the thing that fails here. A direct read closes an `[open]` item; it does not reproduce the guess. (By contrast, §2.13 explicitly lists *"the siege strength shape with archers tripled"* among the claims that **do** hold up — consistent with `Attacker` being correct.)

**Root cause and why I am flagging this for escalation rather than plain rework.** The swap originates *upstream* of this PR: `decompiled-city-capture-resolution.md:24` states `fortification × 150 + loyalty × 250 + <third field> × 200`, T04's corpus fixture `capture.siegeDefenderStrengthFormula` transcribes that string verbatim, and T02's `toy-ruleset.json` names its fields from it. T07 followed its cited sources faithfully. Fixing it properly means renaming `SiegeRules.DefenderFortificationWeight`/`DefenderLoyaltyWeight`, correcting the corpus fixture, and correcting the research report — **all outside T07's Owns list**.

That makes this plan §6.6 case 4: a cited report disagrees with the raw decompilation and the code must pick one. **Orchestrator: please escalate rather than bouncing this straight back to the implementer** — T07 cannot fix it inside its own Owns list. The narrow, in-scope change T07 *can* make is to stop asserting a field identity it cannot honour: either take the three terms positionally (`term150`, `term250`, `term200`) with the ambiguity documented, or block on the upstream correction.

To be explicit, since this is round 1 and I want the disagreement legible if there is a round 2: I am **not** disputing the weights 150/250/200, and I am **not** disputing that T07 should read them from the ruleset. I am disputing only which city field each weight multiplies, and I have given the dump line numbers so the implementer can check the claim directly rather than re-litigating from the report.

---

### Gate 3 — Determinism: **PASS**

No `System.Random`, `DateTime`, `Guid.NewGuid`, `Stopwatch`, or `Environment.TickCount` anywhere in the diff (grepped both `src/` and `tests/`). The only match for "Random" in production is the word inside a doc comment explaining why it is *not* used. The single draw goes through `IRng.NextInt(rules.RandomBandCount)` (`FleetPower.cs:88`), with the count read from the ruleset rather than a literal 4. `FixedDrawRng` is correctly confined to the test project and implements the interface, so production never depends on it. `ArmyPower` and `SiegeStrength` draw no randomness at all. Verified `random(4)` in the original is `FUN_0040284c(4)` (dump 49311), Delphi `Random(N)` over `[0, N)` — matching `IRng.NextInt`'s documented range. Declining to derive a sub-stream inside `Compute`, so the battle resolver controls draw order, is the right call and is documented.

### Gate 4 — Scope: **PASS**

All 9 changed files are inside the Owns list (`src/IC2.Engine/Strength/**`, `tests/IC2.Engine.Tests/Strength/**`); nothing else is touched. **No diff to `docs/build-orchestration-plan.md`**, no original game files, no `assets.local.ini`, no `docs/reports/`. Confirmed by `git diff --name-only main...HEAD`, not by the PR body's table.

The pre-existing `IC2.Data.Tests` corpus-sweep failure is as described — local-asset dependent, T30's territory, untouched here. Not held against this PR.

### Gate 5 — Correctness sweep (`/code-review --effort high`): **FAIL** (one independent latent bug)

The sweep ran over `main...HEAD` and returned 9 findings. Notably, it **independently rederived the gate-2 defender swap from the raw dump without being prompted to** — same two functions, same panel-based field identification, same conclusion. Two independent passes converging on that from the decompilation is why I am treating it as settled fact rather than a reading to argue about.

It confirmed clean, against the dump: `ArmyPower`'s per-unit `/100` then `/80` ordering, the 13-unit Roman hand computation, `FUN_0044A930`'s no-`/100` loop, and `FleetPower`'s `base + draw*(base/10)`.

**New, independent of everything above — please fix this one:**

- **`FleetPower.cs:91` — `RandomBandPercent` is used as a divisor, not a percentage.** The expression is `baseValue / rules.RandomBandPercent`. That is arithmetically correct *only* because the shipped value is 10 and `100 / 10 == 10`. A ruleset that sets `randomBandPercent: 20` intending 20% bands gets **5%** per band — the relationship is inverted. Every other constant in this diff is read from the ruleset in a way that survives being changed; this one silently misbehaves the moment it is. Either compute `baseValue * rules.RandomBandPercent / 100` (and re-pin the truncation test, since the truncation point moves) or rename the field to something like `RandomBandDivisor` so its meaning matches its use. Given that "Nothing here is a C# literal" is the stated design of these three files, I'd take the rename at minimum.

The sweep's remaining findings are the same defender issues already covered in gate 2 (weights swapped, missing `% 100` decode, third field is population), plus the low-severity items folded into "Minor" below.

---

### Hazard compliance (the task entry calls this out twice) — **PASS**

- **No tactical morale.** The per-unit tactical morale array `DAT_004A0350` does not feed any computation in this diff. It is *named* twice (`ArmyPower.cs:35`, `SiegeStrength.cs:36`), in both cases inside a doc comment saying it must **never** appear here, citing `design-audit.md` §2.9. That is compliance, not violation. Both formulas read strategic morale only, and I confirmed against the dump that `FUN_0044a8cc`/`FUN_0044a930` read `(&DAT_0047c1fa)[i * 0x148]` = army record **+14** — exactly the field the hazard requires.
- **`ArmyState.Morale` is read-only here.** Morale is taken as an `int` parameter and used once as a multiplier. No assignment, no clamp, no `Math.Min`/`Math.Max`, no `with`-expression anywhere in the diff. The supply→morale rule is correctly absent — that is T08's. `ArmyPower.cs:48-53` documents the deliberate non-clamping and cites DoD 6's rationale.

Neither automatic-rejection condition is present.

### The deliberate T17 omissions — mostly right, one overstated

Checked against T02's actual merged provenance and T17's task entry rather than the PR body's summary:

- **Owner/allegiance penalty — correctly deferred.** T02's `_provenance` says verbatim *"This is an evidence conflict for whoever implements siege resolution (task T17) to settle, not for this task to pick"*, and T17's entry carries a matching **"Known-open item to record, not resolve"**. Unambiguously right.
- **`× 5/3` bonus — correctly deferred.** T02 flags the branch's guard as unrecovered (`FUN_0044b8d0` is indeed an undecompiled call in the dump). Implementing it would require inventing the gate. Right call.
- **Garrison-troops addend — right outcome, wrong justification.** T02's provenance for `defenderGarrisonTroopDivisor` is a plain `confirmed` with no T17 note, and T17's entry does not mention it either. So the PR's claim that *"per T02's own provenance notes"* all three belong to T17 is accurate for two of three, not three. That said, it is outside DoD 4's literal text and computing it needs nation/army state this pure API does not take, so omitting it is fine — only the stated reason is wrong. Please don't repeat the "T02 assigns it to T17" phrasing; say "needs state outside this function's inputs" instead. **Not a gate failure.**

### Minor, non-blocking

- `SiegeStrengthTests.cs:48-55` — the first half of `Attacker_NoDivisionByHundred_UnlikeArmyPower` is a dead assertion, as its own comment concedes (*"If a /100 were wrongly applied first, this would still be 0"*). It cannot fail for the reason the test name gives. The second half (`:57-61`) carries the whole test. Suggest deleting the first block.
- `ArmyPowerTests.cs:76` — DoD 5 asks for cases where *integer* division differs from *floating-point* division. Cases 2 and 3 assert exactly that via `Math.Round`; this one compares two integer evaluation orders and never pins the float value. It does differ from float (0 vs. 51), so the DoD line is met — but add the matching `Assert.NotEqual((int)Math.Round(8020 / 100.0 / 80.0 * 51), power)` for parity with its siblings.
- `FleetPowerTests.cs:72-73` — `Assert.NotEqual(first, differentSeed)` is deterministic today, but it couples the test to `SplitMix64Rng`'s exact stream. There are only four possible outcomes, so if T03's RNG is ever retuned this becomes a 1-in-4 silent coin flip. Consider asserting the drawn band directly instead.
- **`SiegeStrength.cs:58` — `Attacker` never validates `archerUnitTypeId` against the ruleset**, so it defeats its own stated purpose. The parameter exists (per its doc comment) *"so a custom ruleset that renames its unit types is not silently mishandled"* — but a caller passing `"archer"` instead of `"archers"` gets every archer counted 1× instead of 3×, silently, which is exactly the mishandling it was added to prevent. Compare `ArmyPower.Compute` (`ArmyPower.cs:72-74`), which throws on an unknown unit type. Add `ruleset.UnitTypeById(archerUnitTypeId) ?? throw ...`. The same asymmetry also means a malformed army throws in a field battle but computes quietly in a siege and (via `FleetPower`) a naval battle.
- `FleetPower.cs:102` — `CarriedArmyStrength` is a **record struct**, so `default(CarriedArmyStrength)` is a perfectly valid non-null value that satisfies `carriedArmy is { } army` and reaches `SiegeStrength.Attacker` with a null `Units`, surfacing as `ArgumentNullException(paramName: "units")` — opaque at the `FleetPower` call site. Record-struct equality also silently degrades to reference equality on the `IEnumerable`. A sealed record class, or a null check on `army.Units` with a clearer message, avoids both.
- `FixedDrawRng.cs:30` — `NextInt` ignores `exclusiveUpperBound` entirely and returns `_fixedDraw` unconditionally. `FixedDrawRng(3)` would keep returning 3 even if `RandomBandCount` dropped below 4, quietly making the truncation test assert something the real `IRng` contract forbids. Assert `_fixedDraw < exclusiveUpperBound`.
- `SiegeStrengthTests.cs:67-68` claims the defender weights *"match the T04 corpus's `capture.siegeDefenderStrengthFormula` fixture"*, but that fixture's value is a **prose string**, not numbers — so unlike the `+0x26` weights, nothing actually enforces this cross-check. The comment overstates what the test suite verifies. (This is also, not coincidentally, the fixture carrying the swapped labels.)

---

### Summary

| Gate | Result |
|---|---|
| 1. DoD reproduced | 5 of 6 — DoD 4's fortification/loyalty clause blocked by gate 2 |
| 2. Provenance | **FAIL** — `SiegeStrength.Defender` field/weight mapping contradicts `FUN_0044A98C` |
| 3. Determinism | PASS |
| 4. Scope | PASS |
| 5. Correctness | **FAIL** — `FleetPower.cs:91` uses `RandomBandPercent` as a divisor; plus 8 lower-severity findings |
| Hazards (§2.9 two morales) | PASS — both automatic-rejection conditions absent |

Labelling `status:rework`. The `fleetPower`→`FUN_0044A930` correction and the truncation-order work are genuinely good, and the attacker/defender split should be kept as-is. The defender field mapping is the one thing blocking, and because its fix reaches into T02 and T04, **it wants an orchestrator decision, not a silent round-2 bounce.**

diegoami:
## Round-1 rework — response to the review

Rebased onto `main` at `27d7b41` (T31, "Correct `Ruleset.Siege`'s defender-strength field identities", #48). No change to `Ruleset.Siege` is made by this PR — that was out of Owns and is exactly what T31 fixed.

### Blocking finding (gate 2): `SiegeStrength.Defender`'s field/weight mapping

Rewritten against T31's corrected `SiegeRules` names and `docs/investigations/siege-defender-strength.md`. The formula is now:

```
loyalty × DefenderLoyaltyWeight (150) + finishedFortificationPercent × DefenderFortificationWeight (250) + populationThousands × DefenderPopulationWeight (200)
```

- `fortification`/`loyalty` are no longer swapped against their weights.
- The third term is `populationThousands` (`CityState.PopulationThousands`), a real, named parameter — no parameter or doc comment says "unidentified" or tells a caller to pass 0.
- The fortification word is decoded through `FortificationCode.FinishedPercent` (the guarded `code <= MaxPercent ? code : code % radix` rule already shipped in T02) before being weighted — never the raw stored word, never an unguarded `% 100`.

New tests: `Defender_AppliesLoyaltyFortificationAndPopulationTerms`, `Defender_DecodesInProgressFortificationWord` (130 → 30, not 130), `Defender_FinishedFortificationAtExactlyMaximum_IsNotZeroed` (100 → 100, not 0 — the specific failure mode an unguarded `% 100` causes), `Defender_FormulaShape_MatchesCorpusFixtureWording`.

Not touched, per your own note that this reaches into `Ruleset.Siege`: the `×5/3` bonus and its capital gate (issue #46) and the owner/allegiance `×4/5` penalty (issue #47). Both remain confirmed-but-unimplemented in `SiegeStrength.Defender`, exactly as before.

### Gate 5: `FleetPower.cs:91` — `RandomBandPercent` used as a divisor

Fixed. The bonus is now computed as `baseValue * rules.RandomBandPercent / 100` (one integer division), which is mathematically identical to the old `baseValue / rules.RandomBandPercent` for the shipped value of 10 (10 and 100 share the factor that makes `10x/100` and `x/10` the same rational number, so no existing DoD assertion needed re-pinning) but now scales correctly for any other value. New regression test `Compute_RandomBandPercent_IsAPercentageNotADivisor` builds a ruleset with `RandomBandPercent = 20` and asserts the old (buggy) divisor reading is not what comes out (120 vs. the buggy 105).

### The 8 minor findings

1. **Dead assertion** in `Attacker_NoDivisionByHundred_UnlikeArmyPower` — deleted; kept only the case that can fail for the stated reason.
2. **DoD-5 case 1 missing float-parity assertion** — added, matching its two siblings.
3. **Flaky different-seed assertion** in `Compute_RandomBonus_IsReproducibleUnderTheSameSeed` — replaced with a direct hand-verification against the same seed's own replayed draw (`new SplitMix64Rng(seed).NextInt(...)`), so the test no longer depends on a second, arbitrary seed happening to differ.
4. **`SiegeStrength.Attacker` didn't validate `archerUnitTypeId`** — now throws `ArgumentException` on an unrecognised id, matching `ArmyPower.Compute`. New test `Attacker_UnknownArcherUnitTypeId_Throws`.
5. **`CarriedArmyStrength` record-struct null trap** — changed from `readonly record struct` to `sealed record` (reference type), so `default` can no longer masquerade as a "carrying an army" value with a null `Units`.
6. **`FixedDrawRng.NextInt` ignored `exclusiveUpperBound`** — now throws if the fixed draw doesn't fit the requested range.
7. **Overstated fixture-match claim** — `Defender_FormulaShape_MatchesCorpusFixtureWording` now makes a real `Assert.Contains` check against the (now-corrected) `capture.siegeDefenderStrengthFormula` fixture text, instead of an unchecked comment.
8. **Garrison-troops addend justification overstated** — corrected in the PR body: it's omitted because it needs per-nation recruitment-slot state outside this function's inputs, not because T17 is assigned to resolve an open identity (that's true only for the `×5/3` bonus and the owner/allegiance penalty).

### Re-verified

`dotnet build IC2.sln` — 0 warnings, 0 errors. `dotnet test IC2.sln` — `IC2.Engine.Tests`: 193/193 pass (25 in `Strength`, up from 20). `IC2.Data.Tests` has 3 pre-existing failures from local `saves/` corpus drift on this machine (T30's territory, `src/IC2.Data/**`) — untouched by this branch, confirmed by `git diff --stat main...HEAD`.

`gh issue edit 7 --remove-label status:rework --add-label status:in-review` run after this comment.

diegoami:
## T07 review, round 2 (the last before escalation). Reviewer: Opus / Medium

Reviewed at the new head `2ba90e1`, rebased onto `main` at `27d7b41` (T31 merged), checked out in the main checkout. I checked everything against the code, not the response comment.

`dotnet build IC2.sln`: 0 warnings, 0 errors. `dotnet test tests/IC2.Engine.Tests`: **193 passed, 0 failed**. All 25 Strength tests pass.

**Verdict: approve (`status:approved`).** All five gates pass. The round-1 blocker is fixed properly against T31's corrected names, and so is the `RandomBandPercent` bug. 7 of the 8 minor findings are fully fixed in code; #4 is fixed for the archer id but not for unit ids.

**Three things are not fully right, and none of them blocks:**
1. The seeded reproducibility test draws band 0, so it can't show the draw matters.
2. `Attacker` still doesn't check unit type ids (the other half of round-1 #4).
3. A false comment in `ArmyPower.cs` about which test pins truncation order.

None affects a DoD line or the engine's output, so I'm not holding the merge for a round 3 over them. All three are small fixes (a seed, an id check, a comment). The orchestrator can fold them in at merge or file them as follow-ups.

---

### 1. Round-1 blocking finding: siege-defender field weights. **FIXED**

`SiegeStrength.cs:130-146`:

```csharp
var finishedFortificationPercent = FortificationCode.FinishedPercent(fortificationCode, fortificationOrder);
return (loyalty * rules.DefenderLoyaltyWeight)
     + (finishedFortificationPercent * rules.DefenderFortificationWeight)
     + (populationThousands * rules.DefenderPopulationWeight);
```

Checked against the sharpened DoD 4 one item at a time:

- **The formula is `loyalty×150 + finishedFortificationPercent×250 + populationThousands×200`.** On `main`, T31's `toy-ruleset.json` has `defenderLoyaltyWeight: 150`, `defenderFortificationWeight: 250`, `defenderPopulationWeight: 200`. That matches `FUN_0044a98c` at dump lines 49269-49270 (`DAT_004795a6*0x96 + decoded(DAT_004795aa)*0xfa + DAT_004795ac*200`), with the field identities I established last round from `TInformation_ShowCityDetails`.
- **Every weight is read from `Ruleset.Siege`, with no literals.** `SiegeStrengthTests.cs:103-106` also pins 1500/2500/2000, so a regression back to the swapped assignment fails on absolute values, not just on ratios.
- **Fortification goes through `FortificationCode.FinishedPercent`, never the raw word.** `FinishedPercent` is `code > MaxPercent ? code % radix : code`. With the shipped `fortify` order (max 100, radix 100) that is exactly the dump's `sVar6 < 0x65 ? sVar6 : sVar6 % 100`. Both directions are pinned: `Defender_DecodesInProgressFortificationWord` (130 gives 30) and `Defender_FinishedFortificationAtExactlyMaximum_IsNotZeroed` (100 gives 100, the failure an unguarded `% 100` would cause). Good choice to test the second one.
- **No term is described as unidentified, and no caller is told to pass 0.** The third parameter is `populationThousands`, with a cref to `CityState.PopulationThousands` (which exists on `main`). The word "unidentified" appears three times in the diff: `SiegeStrength.cs:30`, `:35` and `SiegeStrengthTests.cs:114`. Each one describes the *round-1 defect* and says the term is **not** unidentified now. I read that as meeting the rule's intent, which is that no caller is ever told the term is unknown. If the orchestrator reads the rule literally instead, it's a wording fix, not a code fix.
- **The attacker and defender sides stay split, as DoD 4 now requires.** `Attacker` is unchanged apart from the added validation, and it still matches `FUN_0044a930`.

### 2. `FleetPower`'s `RandomBandPercent`. **FIXED**

`FleetPower.cs:107-108`: `bandAmount = (baseValue * rules.RandomBandPercent) / 100; return baseValue + draw * bandAmount;`. For the shipped value of 10 this is exactly `trunc(base/10)` for every integer `base`, negatives included under C# truncation, so no DoD-pinned result moves. That's why DoD 5 case 3 (`17 + 3×1 = 20`) still passes unchanged. The regression test `Compute_RandomBandPercent_IsAPercentageNotADivisor` (`FleetPowerTests.cs:122`) uses the non-default value 20 (base 100, draw 1) and asserts 120, not the old divisor reading's 105. The ruleset is built with `with` expressions, so the shared toy ruleset is not mutated.

### 3. The 8 minor findings: all addressed in code

| # | Round-1 finding | Status at `2ba90e1` |
|---|---|---|
| 1 | Dead first assertion in `Attacker_NoDivisionByHundred_UnlikeArmyPower` | **Fixed.** Deleted. The remaining case (8000 gives 100; a spurious `/100` would give 0) can fail for the reason the test's name gives. |
| 2 | DoD 5 case 1 never compared against floating-point division | **Fixed.** `ArmyPowerTests.cs:99` adds `Assert.NotEqual((int)Math.Round(8020/100.0/80.0*51), power)` (float 51 vs. integer 0). |
| 3 | Brittle different-seed assertion | **Fixed, with a side effect.** The coin-flip assertion is gone, but see the new finding below. |
| 4 | `Attacker` didn't validate `archerUnitTypeId` | **Fixed for the archer id; the unit-id half is still open (see gate 5).** `SiegeStrength.cs:87-91` throws `ArgumentException` for an archer id the ruleset doesn't define. Tested by `Attacker_UnknownArcherUnitTypeId_Throws` with `"archer"`, the exact case from round 1. |
| 5 | `CarriedArmyStrength` record-struct `default` trap | **Fixed.** It is now a `sealed record` (`FleetPower.cs:126`), so `null` is the only way to say "no carried army". |
| 6 | `FixedDrawRng.NextInt` ignored its bound | **Fixed.** It now throws if `_fixedDraw >= exclusiveUpperBound`. |
| 7 | Overstated fixture-match claim | **Fixed.** `Defender_FormulaShape_MatchesCorpusFixtureWording` checks the fixture with `Assert.Contains`. I confirmed that T31 corrected the fixture on `main` to `loyalty*150 + fortification*250 + population*200`. |
| 8 | Garrison-addend justification overstated | **Fixed.** `SiegeStrength.cs:48-51` now gives the right reason: it needs per-nation slot state this function doesn't take. I checked that against the dump: the loop at 49278-49291 walks 40 nation-record slots (`+0x2e8`/`+0x2ea`, stride `0x494`), which the pure function's inputs don't include. The reason holds. |

The two items declared out of scope, the `×5/3` gate and the `×4/5` owner/allegiance penalty, are correctly left to open bugs **#46** ("HighFortificationThreshold is misnamed (tests loyalty, not fortification)") and **#47** ("DefenderOwnerNotAllegiancePenaltyPercent has the wrong formula shape"). Both are open, and both are defects in `Ruleset.Siege`, which is outside T07's Owns list. The sharpened DoD 4 formula is exactly the three-term sum, so leaving them out doesn't create a DoD gap. I grepped: nothing in the diff implements them or works around them (no `HighFortification*`, `*Allegiance*`, `*5/3`, or `<< 2`).

### New finding (not blocking): the seeded test no longer shows the draw matters

`FleetPowerTests.cs:70-88`, `Compute_RandomBonus_IsReproducibleUnderTheSameSeed`. **`SplitMix64Rng(12345).NextInt(4)` returns 0.** I checked this by running the engine DLL through `dotnet fsi`. So `expected = 765 + 0 × 76 = 765`, and a `Compute` that ignored its `IRng` entirely would still pass this test. The remark at `:62-67` says the test proves "the draw genuinely reaches `FleetPower.Compute`". With a zero draw it doesn't.

This doesn't fail DoD 3. Same seed gives the same value, and that is asserted. "Applied through `IRng`" is shown separately by `Compute_RandomBonusTruncationOrder_DiffersFromNaiveRealDivision` (`FixedDrawRng(3)` gives 20) and `Compute_RandomBandPercent_IsAPercentageNotADivisor` (`FixedDrawRng(1)` gives 120): both would fail if `Compute` ignored its `IRng`. But this is the one test that runs a real generator, and fixing finding #3 made it vacuous in a new way.

**One-line fix:** use a seed with a nonzero first draw. Seed `42` draws 2 and seed `1` draws 2 (both checked). Add `Assert.NotEqual(0, expectedDraw);` so a future RNG retune can't quietly make it vacuous again. I'd like this folded in before or at merge. It's not a reason for round 3.

### Gate 5: `/code-review --effort high` over the rebased diff

**PASS.** No behavioural bugs. The sweep checked all four formulas against the dump on its own: `FUN_0044A8CC`, `FUN_0044A930`, `FUN_0044A98C` (guarded decode, then 150/250/200 on loyalty/fortification/population) and `FUN_0044AA54`. It also confirmed that every shipped ruleset value matches. It found two low-severity issues. I checked both myself and neither blocks:

- **`SiegeStrength.cs:96-100`: unit type ids inside `Attacker` are still not checked.** Round 1's #4 was fixed for `archerUnitTypeId` itself, but each *unit's* `UnitTypeId` is still only compared by string, and any id the ruleset doesn't know counts at weight 1. So an army containing a unit typed `"Archers"` or `"bowmen"` throws in `ArmyPower.Compute` but gets a quiet, wrong siege strength here, and the same wrong value flows into `FleetPower`'s carried-army term. That matches the original's bare `type == 2` check, so the numbers are faithful. But the parameter doc at `:74-76` says it validates "for the same reason `ArmyPower.Compute` validates every unit's own type id", and that's only half true. My round-1 wording covered this asymmetry, so this part of #4 is still open. It's robustness, not a DoD issue. The fix is to validate each unit's type with `UnitTypeById(...) ?? throw`, the same as `ArmyPower`, or to narrow the doc comment.
- **`ArmyPower.cs:26-31`: a false claim about what the Roman roster test pins.** The comment says the roster's two heavy-cavalry units make per-unit and sum-first truncation "give genuinely different results". I recomputed it. Per-unit S = 40,506 and sum-first S = 40,507, but **both** give 506 after `/ 80`, so both give 29,854 at morale 59. `Compute_Roman13UnitRoster_MatchesHandComputedValue` would pass even with the `/ 100` wrongly moved outside the loop. Only `Compute_PerUnitTruncation_DiffersFromSumFirstTruncation` (`ArmyPowerTests.cs:76`) guards the order. Someone who trusts this comment could delete that test as redundant and lose the only guard. The comment has been there since round 1 and I missed it then. Fix it by rewording the comment to point at the dedicated test.

---

### Gates

| Gate | Result |
|---|---|
| 1. DoD reproduced | **PASS**, all 6. DoD 1 still gives 29,854 (rechecked). DoD 2: 400 and 444. DoD 3: reproducibility asserted, and the `IRng` path is shown by the fixed-draw tests. DoD 4: both halves, against the sharpened text. DoD 5: three integer-vs-float cases, now all with an explicit float comparison. DoD 6: 51…70 inclusive, strictly monotonic, both bounds asserted. |
| 2. Provenance | **PASS.** Every defender weight comes from T31's corrected `Ruleset.Siege`. The decode uses T02's `FortificationCode`. The `+0x26` weights are still checked against the corpus. No `[designed]` values. The T02 `carriedArmyPowerDivisor` provenance note I flagged in round 1 has already been corrected on `main` by T31, so no bug needs filing for it. |
| 3. Determinism | **PASS.** No `System.Random`, clock, or `Guid` in code. The one draw goes through `IRng.NextInt(RandomBandCount)`. |
| 4. Scope | **PASS.** `git diff --name-only main...HEAD` shows only the 9 files under `src/IC2.Engine/Strength/**` and `tests/IC2.Engine.Tests/Strength/**`. `Ruleset.Siege` and `docs/build-orchestration-plan.md` are untouched. Nothing implements or works around #46/#47. |
| 5. Correctness | **PASS.** Formulas match the dump. Two low-severity non-blocking issues: unit ids in `Attacker` aren't checked, and a false comment in `ArmyPower.cs`. |
| Hazards (§2.9) | **PASS.** Tactical morale is never read in code; `DAT_004A0350` appears only in "must never appear here" doc comments. Strategic morale is taken as an `int`; there are no writes, clamps, or `with` on morale. |

### Cosmetic (not blocking, no round needed)

- `SiegeStrength.cs:7-8` and `:47-48`, and `SiegeStrengthTests.cs:9` and `:78`, still quote DoD 4's **pre-sharpening** wording ("applies the fortification/loyalty term"). The plan now says "applies the city term". `:47-48` says DoD 4 "asks only for the archer-tripling and fortification/loyalty terms", which is no longer the DoD's text. Worth updating whenever the file is next touched.

### Bugs in other merged tasks

None found this round.

PR #77, opened 2026-09-14
T10 News log ring buffer and message catalog

diegoami:
Closes #10

## Task

- **Task**: T10 News log ring buffer and message catalog — [catalogue entry](https://github.com/diegoami/imperial_conquest_2/blob/main/docs/task-catalogue.md#t10-news-log-ring-buffer-and-message-catalog)
- **Branch**: `task/T10-news-log`

## Scope: files touched vs the Owns list

| Changed path | Owns pattern that covers it |
| --- | --- |
| `src/IC2.Engine/News/NewsLogWriter.cs` | `src/IC2.Engine/News/**` |
| `src/IC2.Engine/News/NewsMessageCatalog.cs` | `src/IC2.Engine/News/**` |
| `tests/IC2.Engine.Tests/News/NewsCoverageTests.cs` | `tests/IC2.Engine.Tests/News/**` |
| `tests/IC2.Engine.Tests/News/NewsLogWriterAppendTests.cs` | `tests/IC2.Engine.Tests/News/**` |
| `tests/IC2.Engine.Tests/News/NewsLogWriterFixtures.cs` | `tests/IC2.Engine.Tests/News/**` |
| `tests/IC2.Engine.Tests/News/NewsLogWriterTests.cs` | `tests/IC2.Engine.Tests/News/**` |
| `tests/IC2.Engine.Tests/News/NewsMessageCatalogTests.cs` | `tests/IC2.Engine.Tests/News/**` |
| `tests/IC2.Engine.Tests/News/NewsRingBufferTests.cs` | `tests/IC2.Engine.Tests/News/**` |
| `tests/IC2.Engine.Tests/News/NewsRoundTripTests.cs` | `tests/IC2.Engine.Tests/News/**` |

This is the complete set of files this PR changes relative to `main` (`git diff --name-status origin/main...HEAD`) — unchanged since round 0 of this resumed attempt; round 1 only edited files already in this list.

- [x] Every changed file above is inside the task's declared **Owns** list.
- [x] No change to `docs/task-catalogue.md`, `docs/build-process.md` or any other document, original game files, `assets.local.ini`, `.gitignore`'s exclusion policy, or the research repository's reports.

## DoD evidence

```text
DoD 1: 41 appends leave exactly the 40 newest, in order, oldest evicted.
$ dotnet test tests/IC2.Engine.Tests/IC2.Engine.Tests.csproj --filter "FullyQualifiedName~NewsRingBufferTests"
[...]

PR #86, opened 2026-09-14
T40 Expose the run's published events to systems

### LINKED ISSUE #83: Bug: T03 gives systems no stateless way to read the events published earlier in the run (blocks T10's news writer)
Blocks: T10

**Defect (a gap in merged T03).** A registered system has no stateless way to see the domain events published earlier in the same `TurnCoordinator` run. `SystemContext` exposes only `Events`, a write-only `IEventSink`. `TurnCoordinator.Run` records the run's events in a private `RecordingEventSink` and composes it with the caller's sink in a `CompositeEventSink`, but neither is reachable from a system.

**Why it blocks T10.** T10's DoD 4 needs a writer that turns the turn's news-worthy events into `GameState.NewsLog` entries by the end of the turn, including events from round-scoped phases. T03 requires registered systems to be stateless (`AssemblyScan.Instantiate`), so the writer cannot buffer events itself. After three review rounds, the only ways found both break a contract:
- **Reflection** on `CompositeEventSink._sinks[1]` (R7). It silently misroutes events under a realistic sink setup.
- **A mutable static "current writer"** (the reviewer's prototype). It is shared across registries, games and parallel tests.

Evidence: T10 escalation https://github.com/diegoami/imperial_conquest_2/issues/10#issuecomment-5664096335, and the reviews on PR #77 (https://github.com/diegoami/imperial_conquest_2/pull/77#issuecomment-5663984617, https://github.com/diegoami/imperial_conquest_2/pull/77#issuecomment-5664079375).

**Triage (planner, with the user's decision).** This is fixed by a small correction task, **T40**. It exposes the run's already-published events, each tagged with its phase, on `SystemContext`. T10 is suspended on this bug, then resumes from its branch on Sonnet and uses the seam.

### LINKED ISSUE #84: T40 Expose the run's published events to systems (a T03 seam)
**Spec:** [`docs/task-catalogue.md` → T40](https://github.com/diegoami/imperial_conquest_2/blob/main/docs/task-catalogue.md#t40-expose-the-runs-published-events-to-systems-a-t03-seam). The catalogue entry is the contract: scope, Owns, Done when, hazards, model, reviewer and dependencies. This issue doesn't copy it, so nothing here goes stale.

**Branch:** `task/T40-published-events-seam`
**Closes when merged:** #83 — its PR says `Closes` for each.

This issue's `status:*` label is its status. How a task runs: [build-process.md §4](https://github.com/diegoami/imperial_conquest_2/blob/main/docs/build-process.md#4-the-task-loop). Earlier revisions of this body copied the catalogue entry; see the edit history.

diegoami:
Closes #84
Closes #83

## Task

- **Task**: T40 Expose the run's published events to systems (a T03 seam) — [catalogue entry](https://github.com/diegoami/imperial_conquest_2/blob/main/docs/task-catalogue.md#t40-expose-the-runs-published-events-to-systems-a-t03-seam)
- **Branch**: `task/T40-published-events-seam`

| --- | --- |
| `src/IC2.Engine/Core/Pipeline/SystemContext.cs` | `src/IC2.Engine/Core/Pipeline/SystemContext.cs` |
| `src/IC2.Engine/Core/Pipeline/TurnCoordinator.cs` | `src/IC2.Engine/Core/Pipeline/TurnCoordinator.cs` |
| `src/IC2.Engine/Core/Events/PublishedEvent.cs` (new) | `src/IC2.Engine/Core/Events/**` (additive only) |
| `tests/IC2.Engine.Tests/Core/PublishedEventsFixtures.cs` (new) | `tests/IC2.Engine.Tests/Core/**` (new tests) |
| `tests/IC2.Engine.Tests/Core/PublishedEventsTests.cs` (new) | `tests/IC2.Engine.Tests/Core/**` (new tests) |

- [x] No change to `docs/task-catalogue.md`, `docs/build-process.md` or any other document, original game files, `assets.local.ini`, `.gitignore`'s exclusion policy, or the research repository's reports.

## What this adds

`SystemContext.PublishedEvents` (`ValueList<PublishedEvent>`): everything published earlier in the
current run, in publication order, each tagged with the `TurnPhase` it was published in and the id of
the system that was running at the moment of publication. `PublishedEvent` is a new, additive record in
`Core/Events/`.

`TurnCoordinator.Run` threads a per-run `List<PublishedEvent>` through both `RunPhases` calls (so a
round tick that follows on from a seat's turn still sees that turn's seat-scoped events, and a fresh
`Run()` call always starts the list empty). Each system gets a snapshot of that list taken **before** it
runs (so it never sees anything it is about to publish itself), via a private `TaggingEventSink` that
wraps the run's existing sink (`recorder` + the caller's own sink) and additionally records a tagged copy
of everything published while that system is on the stack. Because the two-argument
`ICommandDispatch.Dispatch(state, command)` overload (`SystemContext.Commands`) and the quarter-boundary
hook (`SystemContext.QuarterBoundary`) both publish through this same per-system sink, events from either
source are automatically tagged and included too — no separate wiring was needed for them. The
three-argument `Dispatch(state, command, events)` overload lets the caller name a different sink and is
therefore never tagged, exactly as it already bypasses `TurnResult.Events` (documented explicitly after
review round 1's N4, below).

This lets T10's news-log writer, registered stateless in `SeatEnd` and `RoundEnd`, read the news-worthy
events of its own scope with no reflection over `CompositeEventSink`'s private sink list (PR #77's R7
finding) and no mutable static (R4), which is exactly bug #83.

## Review round 1

The reviewer set `status:rework` with one blocking finding. Addressed on top of `e1f1b19`, pushed as
`e46a642`:

- **B1 (blocking, gate 1) — fixed.** DoD 3's "a fresh run starts empty, so nothing carries over between
  turns" had no test that could fail against a coordinator whose published-event list outlived one run —
  every existing test built a fresh coordinator and ran it once. Added two tests that reuse the *same*
  coordinator across two runs:
  - `A_second_run_on_the_same_coordinator_does_not_see_the_first_runs_events` (`RunTurn` → `RunTurn`);
  - `A_round_tick_run_on_the_same_coordinator_after_a_turn_does_not_see_the_earlier_runs_events`
    (`RunTurn` with a follow-on → a standalone `RunRoundTick`).

  Confirmed both fail against the exact mutation the review ran (hoisting the `published` list from a
  `Run()` local to a never-cleared coordinator field): both failed with the leaked entries visible in the
  assertion output, then the mutation was reverted (`git diff` against `e1f1b19` after reverting showed no
  residual change to `TurnCoordinator.cs` beyond the two new tests' own commit).
- **N2 (recommended) — addressed.** `The_callers_own_sink_and_TurnResult_Events_are_unaffected_by_the_new_seam`
  now asserts an explicit expected event-kind sequence and full record equality
  (`Assert.Equal(callerPrimary.Events, callerExtra.Events)` / `... result.Events.ToArray()`), and wires the
  caller's own sink as `CompositeEventSink(callerPrimary, callerExtra)` — the shape PR #77's R7 finding
  broke on — instead of a single plain sink.
- **N4 (recommended) — addressed.** Corrected the doc remarks on `SystemContext.PublishedEvents`
  (`SystemContext.cs`) and `PublishedEvent.SystemId` (`PublishedEvent.cs`), which said all three sources
  (direct, command, quarter-boundary) are always tagged. They now say the three-argument
  `Dispatch(state, command, events)` overload is the one exception: it publishes to whatever sink the
  caller names, so it is never tagged and never appears in `PublishedEvents`, consistent with it already
  bypassing `TurnResult.Events`.
- **N1, N3 (non-blocking)** — left for the follow-up, per the coordinator's instruction.

## DoD evidence

```text
DoD 1: "SystemContext exposes the events published earlier in the current run, in publication order,
each tagged with its phase and publishing system id. The list is read-only; a system cannot modify it."
$ dotnet test tests/IC2.Engine.Tests/IC2.Engine.Tests.csproj --filter "FullyQualifiedName=IC2.Engine.Tests.Core.PublishedEventsTests.PublishedEvents_is_exposed_as_a_read_only_immutable_list"
Passed!  - Failed: 0, Passed: 1, Skipped: 0, Total: 1

(Publication order and tagging are also exercised end-to-end by DoD 3's tests below, which assert exact
ordered, phase/system-tagged sequences.)

DoD 2: "It includes events published through command dispatch and quarter-boundary handlers during the
run, since they go through the same sink. One test for each source."
$ dotnet test tests/IC2.Engine.Tests/IC2.Engine.Tests.csproj --filter "FullyQualifiedName=IC2.Engine.Tests.Core.PublishedEventsTests.Events_published_through_a_command_appear_in_the_published_events_view|FullyQualifiedName=IC2.Engine.Tests.Core.PublishedEventsTests.Events_published_through_a_quarter_boundary_handler_appear_in_the_published_events_view"
Passed!  - Failed: 0, Passed: 2, Skipped: 0, Total: 2

DoD 3: "A system in SeatEnd sees the events of SeatStart, Orders and the systems before it in SeatEnd. A
system in RoundEnd sees the round-scoped phases' events. When RunTurn follows on into the round tick, it
also sees the seat-scoped events of the same run, told apart by their phase tag. A fresh run starts
empty, so nothing carries over between turns."
$ dotnet test tests/IC2.Engine.Tests/IC2.Engine.Tests.csproj --filter "FullyQualifiedName=IC2.Engine.Tests.Core.PublishedEventsTests.A_system_in_seat_end_sees_seat_start_orders_and_earlier_seat_end_systems|FullyQualifiedName=IC2.Engine.Tests.Core.PublishedEventsTests.A_system_in_round_end_sees_every_round_scoped_phases_events|FullyQualifiedName=IC2.Engine.Tests.Core.PublishedEventsTests.When_a_seats_turn_follows_on_into_the_round_tick_round_end_also_sees_the_seat_scoped_events|FullyQualifiedName=IC2.Engine.Tests.Core.PublishedEventsTests.A_fresh_run_starts_with_no_published_events|FullyQualifiedName=IC2.Engine.Tests.Core.PublishedEventsTests.A_second_run_on_the_same_coordinator_does_not_see_the_first_runs_events|FullyQualifiedName=IC2.Engine.Tests.Core.PublishedEventsTests.A_round_tick_run_on_the_same_coordinator_after_a_turn_does_not_see_the_earlier_runs_events"
Passed!  - Failed: 0, Passed: 6, Skipped: 0, Total: 6

  (B1 mutation check, not part of the committed suite: hoisting `published` in TurnCoordinator.Run to a
  never-cleared coordinator field and re-running the two new tests above gave
  Failed: 2, Passed: 0, Total: 2 -- both failed with the leaked prior-run entries visible in the
  assertion diff. Mutation reverted; full suite below is against the real implementation.)

DoD 4: "The caller's own sink still receives every event exactly as before, and TurnResult.Events is
unchanged. Existing golden runs keep the same GameStateHash, and every existing test stays green."
$ dotnet test tests/IC2.Engine.Tests/IC2.Engine.Tests.csproj --filter "FullyQualifiedName=IC2.Engine.Tests.Core.PublishedEventsTests.The_callers_own_sink_and_TurnResult_Events_are_unaffected_by_the_new_seam|FullyQualifiedName~TurnPipelineDeterminismTests"
Passed!  - Failed: 0, Passed: 6, Skipped: 0, Total: 6

DoD 5: "dotnet build IC2.sln and dotnet test IC2.sln are green, and git diff --name-only main...HEAD
lists only Owns paths."
$ dotnet build IC2.sln
Build succeeded. 0 Warning(s), 0 Error(s)

$ dotnet test IC2.sln
Passed! - Failed: 0, Passed: 13, Skipped: 65, Total: 78 - IC2.Data.Tests.dll
Passed! - Failed: 0, Passed: 352, Skipped: 0, Total: 352 - IC2.Engine.Tests.dll

$ git diff --name-only origin/main...HEAD
src/IC2.Engine/Core/Events/PublishedEvent.cs
src/IC2.Engine/Core/Pipeline/SystemContext.cs
src/IC2.Engine/Core/Pipeline/TurnCoordinator.cs
tests/IC2.Engine.Tests/Core/PublishedEventsFixtures.cs
tests/IC2.Engine.Tests/Core/PublishedEventsTests.cs
```

## Provenance checklist

- [x] Every constant introduced or changed traces to a `tests/fixtures` entry, a cited research-repo
  report, or a `docs/investigations/` document. — N/A: this task introduces no gameplay constant. It is a
  pure engine seam (a new read-only view and its plumbing); every test fixture value used is either
  derived from state (e.g. `context.State.Calendar.SeasonIndex`) or an arbitrary test-only string literal
  in a test double, exactly like the existing `PipelineFixtures`/`QuarterBoundaryFixtures` T03 left behind.
- [x] Every `[designed]` value states what was searched and came up empty. — N/A: no `[designed]` value in
  this diff.

Values omitted because no source was found: none — N/A, no gameplay numbers in this diff.

## Determinism

- [x] No `System.Random`, wall-clock reads, `Guid.NewGuid`, or order-dependent iteration. The
  `TaggingEventSink` and the `List<PublishedEvent>` it appends to are per-run, local to one
  `TurnCoordinator.Run` call, threaded explicitly the same way `trace` and `signals` already were —
  nothing static, nothing shared across runs (round 1's B1 tests now assert that directly, reusing one
  coordinator across two runs). `PublishedEvents` is `ValueList<PublishedEvent>`, an order-preserving,
  immutable sequence, never a dictionary. The existing determinism guard
  (`DeterminismGuardTests`/`DeterminismScanner`) and the full `TurnPipelineDeterminismTests` suite (byte-
  identical hashes for the same seed, different hashes for a different seed) pass unmodified — see DoD 4/5
  evidence above.

## Docs affected

none — `docs/build-process.md` §2.1 and `docs/task-catalogue.md`'s T10 entry already describe this seam
exactly as built (they were written anticipating T40), so nothing they claim becomes stale.

## Notes for the reviewer

- **Design choice worth double-checking (raised in round 1, and confirmed correct and sufficient there)**:
  an event published through a command a system issues, or through a quarter-boundary handler a system's
  own phase fires, is tagged with the *currently running system's* phase and id — not a separate identity
  for the command handler or the quarter-boundary handler. Command handlers and quarter-boundary handlers
  aren't `IGameSystem`s, so their ids aren't system ids; the DoD's "publishing system id" can only mean the
  system on the stack.
- `SystemContext`'s constructor gained one new parameter (`ValueList<PublishedEvent> publishedEvents`). It
  is `internal` and has exactly one call site (`TurnCoordinator.RunPhases`), so nothing else needed
  updating; confirmed with `grep -rn "new SystemContext("`.
- I found no defect in another task's already-merged code while doing this work.
- Round 1's N1 and N3 are left open for the follow-up, per the coordinator's instruction. P1 and P2 from
  round 1 are for the planner ahead of T10, not findings against this PR.

https://claude.ai/code/session_01MvGDPYqoBFkxJqjJEhzJ7F


diegoami:
## T40 review (reviewer: Opus). Verdict: **status:rework**, one blocking finding (gate 1)

I reviewed PR head `e1f1b19` in a separate worktree, against merge-base `5724a95`. I ran every check myself.

### Gate 1: DoD, reproduced: **FAIL** (B1)

| DoD | Reproduced | Result |
| --- | --- | --- |
| 1. Exposed, ordered, tagged, read-only | `PublishedEvents_is_exposed_as_a_read_only_immutable_list`, plus the exact ordered and tagged sequences in the DoD 3 tests | pass. `ValueList<T>` copies its input (`ValueList.cs:41`) and has no mutator, so a system can't reach the backing `List`. |
| 2. Command and quarter-boundary sources, one test each | `Events_published_through_a_command_…`, `Events_published_through_a_quarter_boundary_handler_…` | pass. Mutation check: routing either `BoundCommandDispatch` or `QuarterBoundaryHook` to the untagged `events` again fails the matching test. |
| 3. SeatEnd / RoundEnd / follow-on / **fresh run starts empty** | the four DoD 3 tests | the first three clauses pass. **The last clause has no check that can fail: B1.** |
| 4. Caller's sink and `TurnResult.Events` unchanged; golden `GameStateHash` unchanged; all tests green | new test + full suite; for the hashes, a base-versus-head probe (below) | pass, but see N1 and N2 |
| 5. build/test green, diff only Owns | `dotnet build IC2.sln`: 0 warnings, 0 errors. `dotnet test IC2.sln`: Data 13 passed / 65 skipped, Engine 350/350. `git diff --name-only main...HEAD`: the 5 files listed in the PR. CI green. | pass |

**B1 (blocking). DoD 3's "A fresh run starts empty, so nothing carries over between turns" is untested.**
`A_fresh_run_starts_with_no_published_events` (`tests/IC2.Engine.Tests/Core/PublishedEventsTests.cs:29-40`) builds a brand-new coordinator and calls `RunTurn` once. Every other new test also makes one run on a fresh coordinator. So nothing ever looks at a *second* run.
- **Mutation I ran.** In `TurnCoordinator.cs:180` I replaced `var published = new List<PublishedEvent>();` with a coordinator-instance field that is never cleared, then ran the full `dotnet test IC2.sln`. Result: **350/350 green**. That mutation is exactly the hazard bug #83 and this task exist to rule out: published-event state that outlives one run.
- **Fix.** Use one coordinator for more than one run and assert on the second run:
  - `RunTurn` → `RunTurn`: the second run's first snapshot is empty, and its SeatEnd reader sees exactly the same sequence as in the first run, not twice as many entries;
  - `RunTurn` with a follow-on → `RunRoundTick`: the round tick's RoundEnd reader sees no seat-scoped entries from the previous run.

**N1 (non-blocking). DoD 4's "existing golden runs keep the same `GameStateHash`" has no committed check that could fail.**
- No test pins a golden hash. `TurnPipelineDeterminismTests`, which the PR body cites as evidence for this clause, compares two runs of the *same* build. A change that moved every hash would still pass it.
- **I reproduced the clause instead.** I wrote a scratch probe (not committed) and ran it at the base and at the head. It dumps, per turn, the `GameStateHash`, every event (`ToString()`) and the trace for:
  - `PipelineFixtures`' script: 64 turns, with dispatch between runs;
  - the full engine registry (`SystemRegistry.FromEngineAssembly()`: T06 calendar and seat rotation, T08 supply/morale and weather): 200 `RunTurn`s including follow-on round ticks, then one `RunRoundTick`;
  - the `quarter-from-phase`, `downstream-consumer` and `commands` groups.
- **Result:** the two outputs are **byte-identical, 821 lines**.
- Non-blocking because it is reproduced, and because the diff can't reach `GameState` by construction. The PR body should still say how the clause was established, rather than citing a test that can't detect it.

**N2 (non-blocking). The DoD 4 test is weaker than its claim.**
`The_callers_own_sink_and_TurnResult_Events_are_unaffected_by_the_new_seam` (`PublishedEventsTests.cs:141-155`) compares the two streams to *each other*, by `Kind` only. If both changed identically, it would still pass.
- Assert `Assert.Equal(sink.Events, result.Events)`. Records have value equality, so this is exact.
- Also assert against an explicit expected sequence.
- Consider a caller sink that is a `CompositeEventSink(recorder, extra)`. That is the wiring that broke PR #77, and asserting it here documents that the seam doesn't depend on it.

### Gate 2: Provenance: **PASS**
There are no gameplay constants in the diff. The test literals are fixture ids, kinds and marker strings only, and the one season index is read from state (`context.State.Calendar.SeasonIndex`). There are no `[designed]` values.

### Gate 3: Determinism: **PASS**
- The diff contains no `System.Random`, wall clock, `Guid.NewGuid`, `Dictionary` or `HashSet`.
- The new state is a per-run, append-only `List`, filled sequentially in pipeline order and handed out as ordered `ValueList` copies.
- RNG stream names and derivation are untouched: `StreamNameFor` and the `SplitMix64Rng.ForStream` call are unchanged.
- The existing seeded-reproducibility tests (`TurnPipelineDeterminismTests`) pass unmodified, and the base-versus-head probe (N1) shows identical hashes and event streams.

### Gate 4: Scope: **PASS**
- All 5 changed files are inside the Owns list:
  - `SystemContext.cs` and `TurnCoordinator.cs` are named in it;
  - `Core/Events/PublishedEvent.cs` is a new file, so the change to `Core/Events/**` is additive;
  - the two files under `tests/IC2.Engine.Tests/Core/**` are new, as required.
- No `docs/*.md` diff.
- "Docs affected: none" is plausible. `build-process.md` §2.1 and the T10 entry already describe this seam, and no document claims `SystemContext` is write-only.

### Gate 5: Correctness (`/code-review --effort high` plus my own read): **PASS** (no blocking bug)

**The flagged interpretation: correct and sufficient.** Events published through a command, or a quarter-boundary handler, are tagged with the phase and id of the *running system*.
- Command handlers and quarter-boundary handlers aren't `IGameSystem`s, so their ids aren't system ids. The DoD's "publishing system id" can only mean the system on the stack.
- The phase tag is what DoD 3 needs, and it is right for both sources:
  - quarter-boundary events land under `CalendarAdvance`, which is round-scoped (`RoundGroup` asserts `CalendarAdvance/test.published.calendar:quarter-handler`);
  - an AI's command events land under `Orders`, which is seat-scoped.

  So a stateless writer that filters on `TurnPhases.ScopeOf(entry.Phase)` gets both right.
- Who emitted an event within that system's call is still recoverable from the event's own type.

**Realistic sink wiring.** The view is built by the coordinator's own per-system `TaggingEventSink`, which wraps `CompositeEventSink(recorder, _sink)`. It never reads the caller's sink. Composing the caller's sink with any number of extra sinks therefore can't misroute or hide anything, which fixes PR #77's R7 failure mode. Nothing is static: the list is a local of `Run`.

**Re-entrancy.**
- A system can't start a nested run.
- `QuarterBoundary.Fire` called several times from one system appends to the same per-system tagging sink.
- `CommandContext` and `QuarterBoundaryContext` don't carry the view, so no handler can observe a list that is still being built.
- An exception discards the per-run list along with the run.

**N3 (non-blocking, efficiency).** `ValueList.From(published)` (`TurnCoordinator.cs:214`) copies the whole list for every system, which is O(systems × events) per run, even when nothing new was published. That is negligible at today's sizes. It could reuse the previous snapshot while `published.Count` is unchanged, since most systems publish nothing.

**N4 (non-blocking, doc precision).** The remarks on `SystemContext.PublishedEvents` (`SystemContext.cs:104-107`) and `PublishedEvent.SystemId` (`PublishedEvent.cs:15-19`) say that events from a command are covered. That holds only for `Commands.Dispatch(state, command)`.
- The three-argument overload, `BoundCommandDispatch.Dispatch(state, command, events)` (`TurnCoordinator.cs:278-279`), publishes to whatever sink the system passes. It bypasses the tagging, just as it already bypasses `TurnResult.Events`, so the behaviour is consistent.
- One clause in the remarks would stop them overclaiming.

**`/code-review --effort high` over the diff.** It found no correctness bug in the engine code. It raised one finding: the fresh-run test runs only once on a new coordinator, so it can't catch carry-over between turns. That is **B1** above, found independently. It also confirmed three things I checked:
- a rejected command's buffered events never reach the tagging sink;
- tagging forwards to the run's sink first, so order and content are unchanged;
- the internal constructor has a single call site.

### For the planner, not findings against this PR

**P1. Commands dispatched outside a run never reach `PublishedEvents`.** This is T10 risk, and a planner question before T10 resumes.
- T03's own usage pattern calls `CommandDispatcher.Dispatch` *between* runs, publishing to the shared caller sink. See:
  - `TurnPipelineDeterminismTests.RunScript` (`tests/IC2.Engine.Tests/Core/TurnPipelineDeterminismTests.cs:164-169`);
  - the `CommandDispatcher` constructor docs (`src/IC2.Engine/Core/Commands/CommandDispatcher.cs:99`: "Pass the same sink the TurnCoordinator was built with").
- A human seat's orders through T23's command layer will very likely take that path. Their events (battles, captures) reach the caller's sink but no run's `PublishedEvents`, so a stateless T10 writer can't see them.
- T40's DoD says "during the run", so this is out of scope here. It does bear on T10's DoD 4 ("a news-worthy event published to the sink during a turn appears … in `GameState.NewsLog` at the end of that turn").

**P2. For T10: the writer has to run last in its phase.** It sees only the systems before it in its own phase, so it needs to be the last system in both `SeatEnd` and `RoundEnd` (for example `Order = int.MaxValue`).
- T06's seat rotation runs in `SeatEnd`, and T12 victory and T17 elimination run in `RoundEnd`.
- With the scope filter, a `SeatEnd` event published after the writer is lost even when the round tick follows on.

diegoami:
## T40 re-review, round 1 (reviewer: Opus). Verdict: **status:approved**

I reviewed head `e46a642` in a fresh worktree and ran every check myself. B1 is fixed, and all five gates pass. The findings below are non-blocking and go to the follow-up. N1 and N3 from round 1 are left for the follow-up by instruction, so I don't count them as findings here.

### Gate 1: DoD, reproduced: **PASS**
- `dotnet build IC2.sln`: 0 warnings, 0 errors. `dotnet test IC2.sln`: Data 13 passed / 65 skipped / 0 failed, Engine **352/352**. CI green on `e46a642`.
- **B1 is fixed.** The two new same-coordinator tests are `A_second_run_on_the_same_coordinator_does_not_see_the_first_runs_events` (`RunTurn` → `RunTurn`) and `A_round_tick_run_on_the_same_coordinator_after_a_turn_does_not_see_the_earlier_runs_events` (a follow-on `RunTurn` → `RunRoundTick`). I re-applied my round-1 mutation locally and then reverted it:
  - **Instance-field leak** (`TurnCoordinator.cs:180`: `published` becomes a never-cleared coordinator field): **both new tests fail** (350/352). On `e1f1b19` the same mutation passed the whole suite.
  - **Static-field leak** (the other form bug #83 warns about): 5 of the 10 `PublishedEventsTests` fail, including both new ones.
- DoD 1, 2 and the first three clauses of DoD 3 are unchanged since round 1 and still pass.
- DoD 4 still holds. `TurnCoordinator.cs` is byte-identical to `e1f1b19`, and the other `src/` changes are `///` comments only, so the round-1 base-versus-head hash probe (821 lines identical) still applies.

### Gate 2: Provenance: **PASS**
No gameplay constants, no `[designed]` values. The rework adds test literals and comments only.

### Gate 3: Determinism: **PASS**
There is no change to engine logic. The diff adds no random, clock, `Guid` or dictionary use. `TurnPipelineDeterminismTests` still passes unmodified.

### Gate 4: Scope: **PASS**
- `git diff --name-only main...HEAD` still lists the same 5 Owns-list files.
- The rework touches only the two doc comments and `PublishedEventsTests.cs`, a new test file.
- No `docs/*.md` diff. "Docs affected: none" is still plausible.

### Gate 5: Correctness (`/code-review --effort high` plus my own read): **PASS** (no behavioural bug)
**N2 is addressed.** The DoD 4 test now:
- asserts an explicit kind sequence;
- checks full record equality across `callerPrimary`, `callerExtra` and `TurnResult.Events`;
- wires the caller's sink as a `CompositeEventSink`.

**N4 is addressed, but the correction overcorrects.** See N5.

**Non-blocking findings (for the follow-up):**

**N5. The new three-argument `Dispatch` wording is inaccurate the other way.** Affected lines: `PublishedEvent.cs:20-23` ("never tagged and never appears here") and `SystemContext.cs:108-111` ("not tagged or recorded here").
- `SystemContext.Events` *is* the per-system `TaggingEventSink` (`TurnCoordinator.cs:220/238`). So `context.Commands.Dispatch(state, cmd, context.Events)`, the obvious call, **is** tagged and **does** appear in `PublishedEvents`.
- I checked this with an uncommitted probe: an Orders system dispatching with the three-argument overload and `context.Events` shows up as `Orders/<system>:<marker>` in the SeatEnd reader's view. `/code-review` confirmed the same independently.
- Accurate wording: events are tagged and recorded **only when the named sink is, or forwards to, `context.Events`**. That is the same condition under which they reach `TurnResult.Events`.
- `/code-review` rated this medium. I'm treating it as non-blocking because it's a doc comment with no behavioural effect: a reader who believes it would only expect *fewer* events than they get.

**N6. Two internal comments still carry the old, unqualified claim.** `TurnCoordinator.cs:216-219` (inline) and `TurnCoordinator.cs:333-339` (`TaggingEventSink` summary) still say any command issued "via `SystemContext.Commands`" routes through the tagging sink, so they now disagree with the corrected public remarks. Fix them with the same wording as N5.

**N7. The B1 test's rationale gives the wrong numbers.** `PublishedEventsTests.cs:181-182` and `:199` say a leaky implementation shows "ten entries" at SeatEnd and "the first run's five" at SeatStart.
- Each SeatGroup run records **6** tagged events: the reader's own snapshot is tagged too.
- So a leak shows **6** at SeatStart and **11** at SeatEnd.
- The assertions are right. Only the comment is wrong.

**N8. The DoD 4 test's composite-sink claim isn't asserted.** `PublishedEventsTests.cs:148` says the composite wiring shows "tagging does not depend on the caller passing a single plain sink". No assertion checks the snapshots' `Seen` contents against an expected list, and every stream is compared only with the others. If tagging broke under a composite sink, all three would match and the test would still pass. Fix: add `Assert.Equal(<the five expected entries>, result.Events.OfType<PublishedEventsSnapshot>().ToArray()[1].Seen.ToArray())`.

### For the planner (unchanged from round 1; not findings against this PR)
- **P1.** Commands dispatched *between* runs, as T03's own `TurnPipelineDeterminismTests.RunScript` does and as T23's human-order path probably will, never reach `PublishedEvents`. A stateless T10 writer can't see them. Triage this before T10 resumes.
- **P2.** T10's writer has to be ordered last in both `SeatEnd` and `RoundEnd`.

